Issue #1, opened 2026-08-18
WSL2 validation: all five profiles pass with three compatibility notes


ThatsR4d:
Thanks for publishing this recipe. I ran it on a single RTX 3090 under WSL2 and reproduced the advertised performance and quality across all five launch profiles. I also found three WSL-specific startup/headroom behaviors that may be worth documenting or addressing separately.

I am opening one report first because the repository does not yet have contribution guidance. I am happy to split this into a small documentation PR and a separate vLLM-level report if that is preferable.

## Environment

- WSL2 kernel: `6.6.87.2-microsoft-standard-WSL2`
- Ubuntu: 24.04.2 LTS
- NVIDIA Windows driver: `591.86`
- GPU: GeForce RTX 3090, 24 GiB, SM86
- Docker Engine / Compose: `29.2.0` / `5.0.2`
- vLLM: `0.27.1`
- vLLM image digest: `sha256:c2f3b1b964e47809b722b5e75b61b1e7b39a50f70388cf2bf2418f16a9f31da2`
- Full five-profile run: recipe commit `e0f5432f5058647e83da67c3d5420e88931f1f78`
- Latest verification run: recipe commit `ee4d48f5282e0ef7077cd44a296773871f7eb795`
- Model revision: `1f05c441c4e64ae0549de44fa9ea5a6d43610314`

The model and compilation caches were on the WSL ext4 filesystem. Only the target RTX 3090 was visible to the serving container.

At `ee4d48f`, the new `verify.sh` completes with zero failures: all five vLLM patches are detected, KVarN imports, every transformed model component is present, and the live completion check passes.

## Validation matrix

Every profile passed authenticated API/chat checks, GPU isolation, two profile-native benchmark runs with zero failed requests, a warm stop/start, and an exact-marker retrieval request near its context limit.

| Profile | Cache tokens | Near-limit retrieval | Output throughput | PPL | GSM8K |
| --- | ---: | ---: | ---: | ---: | ---: |
| `single-long` | 159,326 | 140,087 tokens | 77.06 / 76.59 tok/s | 8.0435 | 96.0% |
| `single-fast` | 81,233 | 59,588 tokens | 89.02 / 88.22 tok/s | 8.0440 | not rerun; same W4A16 weights |
| `single-huge` | 261,818 | 185,088 tokens | 71.24 / 75.05 tok/s | 8.0444 | not rerun; same W4A16 weights |
[...]

PR #4, opened 2026-08-19
docs: add WSL2 validation and memory notes

ThatsR4d:
## Summary

- document an independent WSL2 reproduction at current upstream `e81fa39`
- add the `GPU_UTIL=0.93` fallback for WSL's batch startup free-memory gate
- document a portable, profile-specific procedure for deterministic KV sizing
- record the six-profile cache and throughput matrix, including `KV=int4pth`

The native-Linux defaults remain unchanged. The Marlin-repack allocator
observation from #1 stays separate because it may belong in vLLM.

## Validation

- upstream `verify.sh`: 0 failures
- all six launch configurations: authenticated API/chat and GPU isolation pass
- every recorded benchmark: 0 failed requests
- fast artifact: 8.0940 PPL, 96.0% GSM8K
- `single-fast` greedy C1: 114.17 output tok/s
- `batch-int4pth` C64: 1,043.84 / 1,044.06 output tok/s
- ordinary batch at the WSL-safe 0.93 setting: 201,832 cache tokens, preserving
  the 150k context contract

Related: #1

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,46 @@ tables in this README against the running server (`--prefill` and `--long`
 add the prefill matrix and the long-context rows), and
 `python bench/quality_battery.py <tag>` the perplexity / GSM8K rows.
 
+### WSL2 notes
+
+An independent WSL2 reproduction at `e81fa39` used kernel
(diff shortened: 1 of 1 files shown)


mhenrichsen:
Thanks :)

PR #5, opened 2026-08-19
Windows/WSL2 compatibility: LF endings, stdout-safe path capture, overridable allocator config

LukaszRozmej:
Three fixes from bringing the Docker setup up on Windows 10 + Docker Desktop (WSL2 backend), RTX 3090, Windows driver 610.74. With these, `docker compose --profile single up -d` works end-to-end on WSL2: all `verify.sh` checks pass and single-user mode serves at 76–107 tok/s greedy.

### 1. Force LF line endings (`.gitattributes`)

Cloning on Windows with `core.autocrlf=true` checks out CRLF, and `COPY . .` ships that into the image: bash fails on every script (`set: -\r: invalid option`) and `patch` warns on every hunk. A `* text=auto eol=lf` attribute makes Windows clones build identically to Linux ones.

### 2. vLLM package-path capture breaks under WSL2 (`tail -n1`)

Importing vllm under WSL2 prints `Using 'pin_memory=False' as WSL is detected` **to stdout**, so `SP=$(python -c 'print(os.path.dirname(vllm.__file__))')` captures the warning glued to the path and the Docker build fails at the patch step with `patch: Can't change to directory`. The path is always the last line printed; `| tail -n1` keeps only it. Applied to the Dockerfile, `verify.sh` (path + version capture), and `kvarn/install.sh`. The existing `2>/dev/null` guards don't help because the warning goes to stdout.

### 3. `expandable_segments` crashes Marlin repack on some WSL2 drivers

Both start scripts hard-code `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`. Its CUDA VMM allocation path is rejected by some driver/dxgkrnl combinations: on Windows driver 610.74 (WSL 2.1.5 and 2.7.12 alike) the engine crash-loops with

```
RuntimeError: CUDA driver error: device not ready
```

inside `gptq_marlin_repack` — this looks like the "separate Marlin-repack allocator observation" from issue #1. The `e81fa39` reproduction on driver 591.86 did not hit it, so it appears driver-dependent rather than universal. The scripts now respect a pre-set `PYTORCH_CUDA_ALLOC_CONF`, so affected users put `expandable_segments:False` in `.env`; native Linux keeps the tuned default. Documented as WSL2 note 3 in the README (replacing the issue-#1 caveat paragraph it supersedes).

## Testing

On Windows 10 (19045) + Docker Desktop 29.2.0 + WSL 2.7.12 + RTX 3090 + driver 610.74:

- Docker image builds cleanly; all 8 patches apply; `verify.sh --install` passes at build time.
- Without fix 3: 12+ crash-loop restarts, 100% reproducible at repack. With fix 3: server healthy, full `verify.sh` passes (0 failures), including the live-server chat check.
- Single-user fast-variant mode: 76–107 tok/s greedy across varied 500-token completions.
- Fixes 1–2 are inert on native Linux (LF stays LF; `tail -n1` of single-line output is the same line). Fix 3 keeps `expandable_segments:True` unless the caller sets the variable.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# Everything in this repo executes on Linux (bash, patch, python inside the
+# container image via `COPY . .`), so CRLF working trees break it. Force LF
+# regardless of the cloning machine's core.autocrlf.
+* text=auto eol=lf
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ COPY docker/requirements.txt docker/requirements.txt
 RUN venv/bin/pip install -r docker/requirements.txt
 
 COPY . .
-RUN set -e; SP=$(venv/bin/python -c 'import vllm, os; print(os.path.dirname(vllm.__file__))'); \
+RUN set -e; SP=$(venv/bin/python -c 'import vllm, os; print(os.path.dirname(vllm.__file__))' | tail -n1); \
     for p in patches/*.patch; do echo "== $p"; patch -p1 -d "$SP" < "$p"; done; \
     bash kvarn/install.sh; \
     bash verify.sh --install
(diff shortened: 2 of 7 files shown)


mhenrichsen:
Thanks. Added.

PR #9, opened 2026-08-19
Create wsl2-uva-fallback.patch

silvdoche:
Add WSL2 fallback for UVA buffers in DFlash2 , fails to load without it.

Why this happened
In bare-metal Linux, vLLM's V2 Model Runner uses UVA (Unified Virtual Addressing) zero-copy memory to pass small metadata arrays (a few kilobytes) directly between CPU and GPU. On WSL2, the virtualized DirectX/WDDM driver blocks zero-copy pointer mapping, raising RuntimeError: UVA is not available. The Solution
We can add a simple WSL2 fallback: when UVA is not available, allocate the small buffer directly on the GPU with standard asynchronous memory transfers. The total memory for these metadata buffers is less than 3 MB.

### DIFF
--- a/patches/wsl2-uva-fallback.patch
+++ b/patches/wsl2-uva-fallback.patch
@@ -1,0 +1,33 @@
+cat << 'EOF' > patches/wsl2-uva-fallback.patch
+--- a/v1/worker/gpu/buffer_utils.py
++++ b/v1/worker/gpu/buffer_utils.py
+@@ -43,9 +43,12 @@
+ class UvaBuffer:
+     def __init__(self, size: int | Sequence[int], dtype: torch.dtype):
+-        if not is_uva_available():
+-            raise RuntimeError("UVA is not available")
+         self.cpu = torch.zeros(size, dtype=dtype, device="cpu", pin_memory=True)
+         self.np = self.cpu.numpy()
+-        self.uva = get_accelerator_view_from_cpu_tensor(self.cpu)
++        if is_uva_available():
++            self.uva = get_accelerator_view_from_cpu_tensor(self.cpu)
++            self._has_uva = True
++        else:
++            self.uva = torch.zeros(size, dtype=dtype, device="cuda")
++            self._has_uva = False
+ 
+ 
+@@ -74,3 +77,5 @@
+         dst = buf.cpu if isinstance(x, torch.Tensor) else buf.np
+         n = len(x)
+         dst[:n] = x
++        if not getattr(buf, "_has_uva", True):
++            buf.uva[:n].copy_(buf.cpu[:n], non_blocking=True)
+         return buf.uva[:n]
+@@ -140,4 +145,4 @@
+-        if not uva_instead_of_gpu:
++        if not uva_instead_of_gpu or not is_uva_available():
+             # Create a GPU tensor (default)
+             self.gpu = torch.zeros(size, dtype=dtype, device=device)
+         else:
+EOF
--- a/single-user/start_qwen.sh
+++ b/single-user/start_qwen.sh
@@ -147,6 +147,9 @@ exec venv/bin/vllm serve "$MODEL" \
   --max-num-seqs $MAX_SEQS \
   --api-server-count $API_SERVERS \
   --language-model-only \
+  --enable-auto-tool-choice \
+  --tool-call-parser qwen3_coder \
+  --enable-prefix-caching \
   $ATTN_ARGS \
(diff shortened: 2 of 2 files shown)


cameronr:
In case it's helpful, I ran into either this issue or a similar one? Based on the vLLM docs, I set `VLLM_WSL2_ENABLE_PIN_MEMORY` and that got past the error for me:

```python
# On WSL2 with a compatible kernel (>= 4.19.121), pinned memory is
# supported but disabled by default due to a small performance regression.
# Set to 1 when pinned memory or UVA is required (e.g. CPU offloading
# or v2 model runner).
"VLLM_WSL2_ENABLE_PIN_MEMORY": lambda: bool(
int(os.getenv("VLLM_WSL2_ENABLE_PIN_MEMORY", "0"))  
),
```

https://docs.vllm.ai/en/stable/configuration/env_vars/

mhenrichsen:
Closing this, but the bug report inside it was real and it is now documented — thank you for finding it, and @cameronr for the comment that turned out to be the answer.

**@cameronr is right, and I traced the chain to be sure.** In 0.27.1:

```
UvaBuffer.__init__  ->  is_uva_available()                       platform_utils.py:51
                        -> is_pin_memory_available()             "UVA requires pinned memory"
                           -> cuda.py:303-306, on WSL2:  return envs.VLLM_WSL2_ENABLE_PIN_MEMORY
                                                         (envs.py:292, default False)
```

So `RuntimeError: UVA is not available` on WSL2 is **a default, not a capability**. vLLM's own comment two lines up says it: *"On compatible WSL2 kernels, pinned memory is supported but disabled by default"* — disabled for a small performance regression, not because it does not work. `VLLM_WSL2_ENABLE_PIN_MEMORY=1` flips it and the normal zero-copy path runs.

That is why I would rather not carry a patch here: it adds a code path to work around a one-line default, and it is a code path that only WSL users would ever execute, i.e. the configuration I cannot test.

It is now the **first** WSL item in [`docs/docker.md`](https://github.com/syv-ai/qwen38-27b-rtx3090/blob/main/docs/docker.md), flagged as a hard abort rather than a tuning question, and scoped correctly — `SPEC=dflash2` forces the V2 model runner on *every* `CTX` profile, not only `CTX=huge`, so `CTX=fast` needs it too. One trap worth repeating from [#25](https://github.com/syv-ai/qwen38-27b-rtx3090/issues/25): check the spelling. `VLLM_WSL_PIN_MEMORY` is not a vLLM variable and reads as a silent no-op.

Three things about the patch itself, in case you want to take another run at it:

**1. The file cannot apply — it contains the shell command that was meant to create it.** Its first line is `cat << 'EOF' > patches/wsl2-uva-fallback.patch` and its last is `EOF`, so the heredoc got committed instead of executed. `git apply --check` stops at `patch fragment without header at line 27`, and `patch -p1 --dry-run` fails all three hunks. `verify.sh` would report this as a missing patch rather than a broken one, which is the worse of the two.

**2. It keeps `pin_memory=True` on the CPU tensor.** The allocation the platform gate is about is still requested; only the view is changed. So even applied, it is treating the symptom one level below where the decision is made.

**3. Hunk 3 changes what `uva_instead_of_gpu` means.** `if not uva_instead_of_gpu or not is_uva_available()` sends the fallback down `self.gpu = torch.zeros(..., device=device)`, and that flag exists precisely so a large, rarely-touched tensor stays *off* the GPU (`buffer_utils.py:139-142`: *"For a large but not-frequently-accessed tensor, we can use UVA instead of GPU to save GPU memory"*). The fallback would silently move it back into VRAM, on the platform with the least headroom to spare.

**Separately, the second file in this PR should not be here at all.** It hardcodes three flags into `single-user/start_qwen.sh`:

```
+  --enable-auto-tool-choice \
+  --tool-call-parser qwen3_coder \
+  --enable-prefix-caching \
```

Those are knobs now — `TOOLS=1` (default on, with `TOOL_PARSER`) and `PREFIX_CACHE`. The prefix-caching one would be a real regression rather than a redundancy: the `PREFIX_CACHE=1` branch adds `--mamba-cache-mode align` alongside it (`start_qwen.sh:286`), and on a hybrid model bare `--enable-prefix-caching` without the aligned mamba mode is not the same feature. Adding it to the exec line makes it unconditional *and* strips the half that makes it correct.

If you do want to keep a local fallback for a machine where the env var is not enough, the shape I would take upstream is a change to `is_pin_memory_available()`'s WSL branch rather than to `UvaBuffer` — one decision point instead of three, and it would be testable by anyone.

PR #23, opened 2026-08-21
Serve images with VISION=1, instead of fighting --language-model-only from EXTRA_ARGS

PassivePicasso:
## The problem

`--language-model-only` is hardcoded in both exec lines, so the only way to serve
images is to pass `--no-language-model-only` through `EXTRA_ARGS` and rely on
argparse taking the last flag. That works today only because `${EXTRA_ARGS}` is
expanded last — this update moved `${TOOL_ARGS}` in one line above it.

The failure mode when it does break is the bad kind: nothing errors. Images are
still accepted, still counted as prompt tokens, and the model answers from
placeholder embeddings — which reads as the model being bad at vision rather than
as a server that quietly dropped the tower.

## The change

One knob, `VISION`, in both start scripts, in the same shape as `TOOLS`:

- `VISION=0` (default) → `--language-model-only`, exactly as now. The command line
  is unchanged for anyone who does not set it.
- `VISION=1` → `--limit-mm-per-prompt {"image":{"count":1}}` plus a 2048-image-token
  pixel cap.

Only `--language-model-only` gets a knob, because it is the only one that cannot be
countered from `EXTRA_ARGS`. The two flags `VISION=1` adds have no competing
hardcoded flag and stay overridable there, since `${EXTRA_ARGS}` is expanded after
them. The pixel cap is shipped rather than left to the processor default because
vLLM profiles the encoder at the largest image it will accept, and that peak comes
out of the KV pool.

## Gotcha 9's 2.7 GB does not hold on these checkpoints

The tower is **0.858 GiB**, confirmed two independent ways:

- `model.visual.*` sums to 0.858 GiB of BF16, read from the safetensors headers, in
  both `Qwen3.8-27B-W4A16-AutoRound` and the `-fast` variant.
- A runtime A/B on one server, one variable changed (`CTX=long`, `PREFIX_CACHE=1`,
  `SPEC=mtp`, fast variant, RTX 3090): model loading 15.13 GiB with the tower against
  14.26 without. Non-weight overhead comes to 0.42 GiB and 0.41 GiB respectively, so
  the delta is the tower and not something else.

Quantization is not the explanation — the tower is BF16 in both dirs. Where the
2.7 GB came from I could not work out, so the entry now gives the measured figure and
says as much.

I am also not claiming the KV pool difference means anything. It came out within 0.4%
across that pair (183,673 against 184,438 tokens at 150k/fp8), but the profiled
activation peak differed by 0.87 GiB between the two starts — the same run-to-run
swing the V2 runner shows — which is larger than the effect being measured.

## Testing

- `VISION=0`: loads 14.26 GiB and logs `running in text-only mode` — the pre-change
  path, unchanged.
- `VISION=1`: 15.13 GiB and 183,673 tokens, identical to the same server run with the
  old `EXTRA_ARGS` trick, so the knob is behaviour-preserving rather than a new path.
- OCR round-trip through `/v1/messages` (rendered text in a PNG, transcribed back):
  3/3 exact.
- Tool calling unaffected alongside it (`qwen3_coder`, real `tool_use` block).
- C1 decode 84.4 tok/s against 84.3 before the change.
- `bash -n` on both scripts; `docker compose build` clean, all 14 patches PASS.

Docs: gotcha 9, a `VISION` row in the single-user Knobs table, the compose knob list,
and the batch script header.

---

NOTE:
To be clear, while I looked at these changes and it seemed reasonable to me and I'm a software developer, this is not an area where I have expertise, and I did not make these changes myself.
I did not run extensive benchmarking, as this is a knob that defaults to off, it maintains existing functionality, but creates an easy access point to enable a feature some people might want.

### DIFF
--- a/batch/start_qwen.sh
+++ b/batch/start_qwen.sh
@@ -13,7 +13,8 @@
 #    change for +2.2% perplexity. Needs both marlin patches from patches/.
 #    INT8_LAYERS=gate_up is the gentler variant (+0.9% PPL, ~+15%);
 #    INT8_ACT= (empty) turns it off entirely (pure W4A16, quality-neutral).
-#  - --language-model-only skips the vision tower entirely (~2.7 GB saved)
+#  - --language-model-only skips the vision tower entirely (0.858 GiB on this
+#    checkpoint); VISION=1 keeps it for a client that sends images
 #  - expandable_segments is required: the DeltaNet prefill kernels allocate
 #    transient workspace and fragment the allocator, OOMs at util >= 0.978 without it
 #  - gpu-memory-utilization 0.972 is the sweet spot on a headless box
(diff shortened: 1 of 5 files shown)


mhenrichsen:
Merged. I re-verified this on current `main` rather than on your branch, because the exec line has moved since you opened it, and everything you claimed holds.

**Gotcha 9's 2.7 GB was wrong and your 0.858 GiB is right**, confirmed the same two ways you used, independently:

```
Qwen3.8-27B-W4A16-AutoRound:      333 visual tensors, 0.858 GiB, dtypes=['BF16']
Qwen3.8-27B-W4A16-AutoRound-fast: 333 visual tensors, 0.858 GiB, dtypes=['BF16']
```

and a runtime A/B on the merged tree, one variable, `SPEC=dflash2 CTX=fast PREFIX_CACHE=1`, fresh boot each:

| | model loading | KV pool | `text-only mode` logged | image request |
|---|---|---|---|---|
| `VISION=1` | **15.88 GiB** | 68,605 tokens | no | answered from the pixels |
| `VISION=0` | **15.02 GiB** | 68,605 tokens | yes | clean `400`, "At most 0 image(s)" |

0.86 GiB apart, which is your number. On this pair the KV pool came out **identical** rather than within 0.4% — one fewer thing to caveat, though your caution about the activation-peak swing was the right call to make.

The OCR check was the one I most wanted, since the regression you are guarding against is the silent one. I rendered a word into a PNG that appears nowhere in the prompt and asked for it back: `VISION=1` returns `SALAMANDER`, `VISION=0` refuses the request outright with a 400 rather than answering from placeholder embeddings. So the shipped default fails loudly; it is specifically the `--no-language-model-only`-through-`EXTRA_ARGS` route that could go quiet, which is exactly the argument in your description.

Two mechanical things I checked because they would fail silently:

- `${VISION_ARGS}` word-splits into exactly the four intended argv entries under `bash`, and the JSON braces survive — brace expansion runs *before* parameter expansion, so `{"image":{"count":1}}` is not expanded. (It does **not** split under `zsh`, which does not word-split unquoted expansions, but that is true of `$KV_ARGS`, `$ATTN_ARGS` and `${EXTRA_ARGS}` already and the scripts are `#!/bin/bash`.)
- Clean merge into `main` and `bash -n` on both scripts after it.

The reasoning I liked most is why the pixel cap ships rather than being left to the processor default — vLLM profiles the encoder at the largest image it will accept, and that peak comes out of the KV pool, so leaving it unset makes the pool depend on a number nobody set deliberately. That is the kind of thing that gets discovered a year later.

On your note: you were right to flag that you did not write the change yourself, and it did not cost the PR anything — the claims were checkable, and they checked out. Thank you.

PR #24, opened 2026-08-22
Add mode-aware vision tower placement

hongguifeng:
First, thank you to the original author and contributors for this project. The carefully engineered optimizations in this repository deliver a major leap in Qwen3.8 decoding speed on a single RTX 3090—particularly through speculative decoding, targeted quantization, optimized attention and KV-cache paths, and GPU kernel tuning—making low-latency local inference genuinely practical on this hardware. This PR builds on that high-performance text-inference foundation by enabling the checkpoint's existing vision capability while preserving the text-only default and balancing image latency, VRAM use, and context capacity across operating modes.

## Summary

This PR adds opt-in, mode-aware placement of the Qwen3.8 vision tower for the single-user server:

- `VISION=0` (default): retain the current text-only behavior via `--language-model-only`.
- `VISION=gpu`: keep the vision tower in VRAM for lower image TTFT.
- `VISION=uva`: keep the 0.86 GiB BF16 vision tower in pinned CPU RAM and let the GPU access it through vLLM's UVA zero-copy path during image encoder cache misses.
- `VISION=auto`: use GPU placement for `CTX=long`, and UVA placement for `CTX=fast` / `CTX=huge`.

The default remains `VISION=0`, so existing users retain the current memory use, context limits, and text-only behavior unless vision is explicitly enabled.

## Why a vLLM patch is needed

vLLM 0.27.1 already provides `UVAOffloader`, `--cpu-offload-gb`, and `--cpu-offload-params`, but its normal offloader hook only wraps modules constructed through `make_layers()` (the decoder layers). The Qwen3.5/Qwen3.8 vision tower is constructed directly as one `Qwen3_VisionTransformer` module, and parameter matching is performed against names relative to the wrapped module. Therefore `--cpu-offload-params visual` does not reach the tower, while non-selective offloading would move decoder weights that are read on every decode step.

`patches/qwen3_5-visual-uva.patch` routes the dense `Qwen3_5ForConditionalGeneration.visual` module through the existing `UVAOffloader` after weight loading. It also:

- activates only for the exact visual-only UVA configuration;
- validates that `--cpu-offload-gb` is large enough for the complete tower, avoiding partial offload;
- is idempotent if weight loading is invoked more than once;
- leaves decoder weights untouched;
- is a no-op for text-only and GPU-vision configurations.

The CUDA view retains ownership of the pinned CPU tensor in vLLM's C++ implementation, so the CPU storage remains alive after the local offloader helper returns.

## Placement policy

The `auto` policy reflects measured capacity and latency trade-offs:

- **Fast / DFlash2 / BF16 KV:** UVA preserves the full 65,536-token limit. The measured pool remained 68,605 tokens.
- **Long / MTP / FP8 KV:** GPU placement reduced the pool from approximately 210k tokens (text-only) to approximately 182-183k, but still preserved the full 150,000-token request limit while reducing image TTFT.
- **Huge / KVarN:** UVA prioritizes the context capacity that this mode exists to provide.

On the tested platform, for a 768x512 image with thinking disabled and a one-token streamed response, warm TTFT was approximately:

- GPU tower: 0.45 s
- UVA tower: 1.42 s

The approximately 0.97 s difference is image/interconnect/platform dependent. Text-only requests do not execute the vision encoder, and the selective UVA path does not add vision-weight PCIe reads to text decoding. The main measurable cost of GPU vision placement is reduced KV-cache capacity. Fresh WSL2 starts showed noticeable runtime and speculative-acceptance variance, so decode tok/s should be compared with matched warm runs and per-step timing rather than with a single absolute number.

## Multimodal safety limits

The checkpoint's processor permits very large inputs (16.7M image pixels and up to 768 video frames), which is unsafe as a default on a 24 GiB card. Vision-enabled starts therefore use:

- `max_pixels=1,048,576` (about 1,024 visual tokens per image);
- up to four images per prompt by default (`VISION_IMAGES=4`);
- video disabled by default (`VISION_VIDEOS=0`).

A 2048x2048 input image tokenized to about 1,080 total prompt tokens, confirming that the 1M-pixel resize cap was applied. Four images were accepted; a fifth image was rejected with the expected HTTP 400 limit error.

### Test platform

- GPU: NVIDIA GeForce RTX 3090, 24,576 MiB
- NVIDIA driver: 610.88
- Host: Ubuntu 22.04.5 LTS under WSL2
- Kernel: `6.18.33.2-microsoft-standard-WSL2`
- Docker: 29.1.3
- Docker Compose: 2.40.3
- vLLM: 0.27.1
- PyTorch: 2.13.0+cu130
- CUDA runtime/toolkit: 13.0 / nvcc 13.0.88
- Model: the repository's `Qwen3.8-27B-W4A16-AutoRound-fast` variant

### Build and static validation

- Built the Docker image with every `patches/*.patch` applied in repository order.
- `verify.sh --install`: 0 failures.
- Applied all patches to a clean vLLM 0.27.1 source tree.
- Compiled the patched `qwen3_5.py` with `py_compile`.
- `patches/_check_applied.py` passed for the new patch.
- `bash -n single-user/start_qwen.sh` passed.
- `git diff --check` passed.

### Functional vision validation

A generated 768x512 test image contained:

- 2 red squares;
- 3 blue circles;
- 1 green triangle.

Both GPU and UVA modes returned the correct result (`red=2, blue=3, green=1`) through the OpenAI-compatible `image_url` chat API.

### Decode A/B testing

To measure whether vision placement changes pure-text decoding, the same server was restarted for each configuration and tested without sending any images. Each case used greedy sampling, disabled thinking, a fixed 768-token output, streaming timing from the first output token onward, a complete warm-up, and three measured repetitions. We tested 4K single-stream, 4K four-stream, and 50K single-stream prompts for Fast; Long additionally used 100K single-stream prompts. Speculative-decoding acceptance metrics were recorded alongside decode timing.

The most stable comparison was Fast 4K single-stream: `VISION=0` measured 27.96 ms per speculative step and `VISION=uva` measured 27.89 ms per step. The corresponding effective rates were 101.6 and 106.2 tok/s, but UVA accepted 2.969 tokens per step versus 2.856 for text-only, so the tok/s difference came from speculative acceptance rather than a slower or faster decoder step. A second matched run at 50K measured 39.63 ms/step for text-only and 35.63 ms/step for UVA, again with different acceptance (3.084 versus 3.448 tokens/step).

Long-mode absolute tok/s varied more across fresh WSL2 starts: GPU vision measured 61.5/52.0/50.0 tok/s at 4K/50K/100K, while text-only measured 65.1/55.4/60.0 tok/s and UVA measured 78.7/62.3/77.5 tok/s. Since UVA cannot be faster through a vision-weight path that pure-text requests never execute, these results show that startup state, GPU clocks, asynchronous scheduling, and speculative acceptance dominate the cross-restart variance. The reliable result is that the vision tower is absent from the pure-text decode path; GPU placement costs VRAM and KV capacity, while UVA preserves nearly all of that capacity.

### Fast mode: UVA vision plus near-limit context

Observed at startup:

- CPU-offloaded vision parameters: 0.86 GiB
- GPU KV cache: 68,605 tokens
- configured maximum: 65,536 tokens

A combined image + long-text request completed successfully with:

- prompt tokens: 65,304
- total tokens: 65,411
- correct image answer
- HTTP 200 / normal stop

### Long mode: GPU vision plus near-limit context

Observed at startup:

- GPU KV cache: approximately 182-183k tokens
- configured maximum: 150,000 tokens

A combined image + long-text request completed successfully with:

- prompt tokens: 149,654
- total tokens: 149,741
- correct image answer
- HTTP 200 / normal stop
- no CUDA OOM or engine failure after completion

The pure-text decode A/B test above found no measurable Fast UVA decoder-step penalty and did not establish a fixed Long-mode tok/s penalty attributable to the vision tower. The Long-mode trade-off that was consistently measurable was memory capacity.

For comparison, measured long-mode pools on repeated starts were approximately:

- `VISION=0`: 210k
- `VISION=uva`: 207k
- `VISION=gpu`: 182-183k

All three retained the full 150,000-token per-request limit.

## Scope and limitations

- The vLLM patch targets the dense `Qwen3_5ForConditionalGeneration` used by this 27B checkpoint. The MoE conditional-generation class is not patched.
- The patch is written and validated against vLLM 0.27.1, matching this repository.
- UVA requires pinned host memory. On WSL2, `VLLM_WSL2_ENABLE_PIN_MEMORY=1` is required; Windows/WDDM pinned-memory limits still apply.
- Video remains disabled by default. It should only be enabled together with explicit, conservative frame and size limits.

## AI contribution disclosure

The solution design, implementation, code changes, review, and the validation described in this PR were primarily completed by **GPT-5.6-sol**, with the user providing the requirements, access to the RTX 3090 test platform, and supervision of the work.

### DIFF
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -340,3 +340,19 @@ Things that each cost us hours, in rough order of pain. Worth skimming before yo
     instead of length; `bench/labd_bench.py` sends two warm-ups on `doc[:4000]`,
     which arms the trigger for everything after it. `bench/bugb_sweep.py` prints
     the `mod 128` column for this.
+
+14. **The vision tower cannot be selected by vLLM's built-in offloader.**
+    `--cpu-offload-params` only matches modules built through `make_layers()`
+    (the decoder), on names *relative to the wrapped module* — the Qwen3.5
+    tower is one module constructed directly in `__init__` (names
+    `blocks.N` / `merger` / `patch_embed`), so it can never match `visual`
+    from the CLI, and a bare `--cpu-offload-gb N` takes the first N GB of
+    decoder layers — exactly the weights read on every decode step.
+    `patches/qwen3_5-visual-uva.patch` routes the tower through the same
+    UVAOffloader after `load_weights` (what `VISION=uva` in the single-user
+    script uses); it is a no-op on pure-text servers. Related trap: any
+    vision-enabled start must cap `max_pixels` (the script sets 1M) and the
+    modality counts (four images, video disabled by default). The checkpoint's
+    16.7M-pixel / 768-frame defaults make profiling and request memory unsafe
+    on 24 GB; enable video only with an explicit small `VISION_VIDEOS` plus
+    frame/size limits.
--- a/patches/qwen3_5-visual-uva.patch
+++ b/patches/qwen3_5-visual-uva.patch
@@ -1,0 +1,123 @@
+CPU-resident vision tower for Qwen3.8-27B (vLLM 0.27.1).
+
+Qwen3.8 is a VL checkpoint: the Qwen3_5ForConditionalGeneration target model
+carries a 921 MB BF16 vision tower (27-layer ViT, hidden 1152).
+--language-model-only drops it entirely; without this patch there is no way
+to keep the *tower* out of the GPU selectively. vLLM's built-in offloader
+(UVAOffloader: --offload-backend uva --cpu-offload-gb N
+--cpu-offload-params <segments>) only wraps modules built through
+make_layers() (the decoder), and its segment matching runs against names
+*relative to the wrapped module* - so the tower, one module constructed
+directly in __init__ with names like blocks.N / merger / patch_embed, can
+never match "visual" from the CLI. (The non-selective budget mode is worse:
+it takes the first N GB of *decoder* layers, i.e. exactly the weights read
+on every decode step.)
+
+The patch routes the tower through the same UVAOffloader once
+load_weights() is done: weights move to pinned CPU memory and the GPU
+dereferences them over PCIe (zero-copy UVA view) whenever the encoder runs.
+The tower only executes during image prefill, never per decode step. On this
+3090/WSL2 box a 768x512 image's warm TTFT was ~1.42 s with UVA vs ~0.45 s
+with GPU weights (image/interconnect dependent); the KV pool keeps the ~0.9
+GB the tower weights would otherwise occupy, while encoder profile/cache
+still consume some GPU memory. That is what lets CTX=fast
+(bf16 KV, 64 KB per token) run with vision and still hold the full 65,536
+context: the measured pool is 72,475 tokens, and the tower on the GPU would
+shrink it to ~56k, which does not fit 64k.
+
+Only active when the server is started with UVA offloading and exactly
+--cpu-offload-params visual (--offload-backend uva --cpu-offload-gb > 0); on
+a pure-text start it is a no-op, and visual matches no decoder parameter, so the
+built-in offloader's LLM side stays empty either way. If the UVA view is
+unavailable (e.g. VLLM_WSL2_ENABLE_PIN_MEMORY unset, or
+VLLM_WEIGHT_OFFLOADING_DISABLE_UVA=1), UVAOffloader falls back to copying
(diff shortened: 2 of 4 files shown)


hongguifeng:
Testing reveals a slight performance degradation after enabling the vision module, which requires further investigation.

PR #36, opened 2026-08-24
Offload the vision tower to host RAM, so VISION=1 works with SPEC=dflash2

danieldonaldson:
**edit to add** - Thanks to the maintainer of this. It's been working amazingly on my 3090! I have a bit higher VRAM requirements due to running Omarchy and a browser (otherwise my computer is useless) so the context is slightly reduced. This has been a real pleasure to use!

`VISION=1` (#23) works on `SPEC=mtp` and does not work on `SPEC=dflash2` — the mode
the README recommends to anyone who owns the card. On a 3090 at 250 W, the dflash2
path dies in graph capture:

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 960.00 MiB.
GPU 0 has a total capacity of 23.55 GiB of which 787.50 MiB is free.
  spec_decode_attn.py:184  ->  self.part_o = torch.empty(n, head_dim, ...)
```

Reproduce with `VISION=1 SPEC=dflash2 bash single-user/start_qwen.sh` on 24 GB.

### Why

The dflash2 pool is pinned by bytes (`KV_MEM`), so the 0.858 GiB tower cannot come
out of the KV cache — it comes out of the ~1.1 GiB transient margin the `KV_MEM`
comment sizes, and it is 0.85 of it. The split-KV verify buffer then has nowhere to go.

`--cpu-offload-gb` cannot fix this. vLLM's offloader is installed in exactly one
place, `make_layers()` (`model_executor/models/utils.py`), which the language decoder
layers use and the ViT does not — `Qwen3_VisionTransformer` builds `self.blocks` as a
plain `nn.ModuleList`. Setting `--cpu-offload-gb` spends the budget on language layers
instead, which puts PCIe in the decode hot loop.

### The change

`patches/vision-tower-cpu-offload.patch` routes the tower's `blocks` and `merger`
(870 of its 878.8 MiB) through their own `UVAOffloader` — dedicated, so it never
competes with `--cpu-offload-gb` — with zero-copy off, leaving the bulk-copy path:
each module is copied H2D once per forward and freed. `patch_embed` and `pos_embed`
(8.5 MiB) stay resident, so `visual.device` / `visual.dtype` still report cuda.

`VISION_OFFLOAD` defaults to 1 in both start scripts, because the alternative is a
config that hard-OOMs and the cost is 36 ms per image.

### Measured

Isolated tower, RTX 3090 at PCIe 4.0 x16, one 8192-patch image (the 2048-image-token
cap the start scripts ship), median of 10 forwards:

| `VLLM_VISION_CPU_OFFLOAD_GB` | weights on GPU | peak alloc | forward |
|---|---|---|---|
| 0 (stock) | 891.3 MiB | 1160.5 MiB | 296 ms |
| 1 | 9.0 MiB | 308.2 MiB | 333 ms |

Output is **bit-exact** against the resident tower (same `state_dict`, `torch.equal`).

End to end, `VISION=1 SPEC=dflash2`: boots at the full 69,758-token pool at `CTX=fast`,
and an OCR round-trip reads the pixels.

### Two notes for review

**Do not use vLLM's UVA zero-copy path here.** Same memory, **3327 ms** per image,
because the GEMMs then re-read operand tiles over PCIe in the inner loop (~290 MB/s
effective). The patch forces the copy path on its own instance rather than via
`VLLM_WEIGHT_OFFLOADING_DISABLE_UVA`, so `--cpu-offload-gb` elsewhere is unaffected.

**The `SPEC=mtp` half is a null result and is written as one.** That path boots either
way, and its pool is profiling-sized, so what the tower costs there is buried in the
±0.87 GiB profiling swing gotcha 9 already warns about — 79,271 against 80,055 tokens
across a `VISION=0`/`VISION=1` pair, 1%, i.e. noise. No context win is claimed.

`verify.sh` passes with 0 failures and lists the patch as applied.

### DIFF
--- a/batch/start_qwen.sh
+++ b/batch/start_qwen.sh
@@ -106,6 +106,21 @@ TOOL_ARGS=$([ "${TOOLS:-1}" = 1 ] && echo --enable-auto-tool-choice --tool-call-
 # 2097152 px = 2048 image tokens.
 if [ "${VISION:-0}" = 1 ]; then
   VISION_ARGS='--limit-mm-per-prompt {"image":{"count":1}} --mm-processor-kwargs {"size":{"shortest_edge":65536,"longest_edge":2097152}}'
+  # VISION_OFFLOAD keeps the tower's weights in pinned host RAM and copies each module to
+  # the GPU for the duration of its own forward (patches/vision-tower-cpu-offload.patch).
+  # It defaults ON, because on 24 GB SPEC=dflash2 + VISION=1 does not boot without it:
+  # the tower is 0.85 GiB of the ~1.1 GiB transient margin the KV_MEM comment sizes, and
+  # graph capture then dies allocating the split-KV verify buffer --
+  #   torch.OutOfMemoryError: Tried to allocate 960.00 MiB ... 787.50 MiB is free
+  #     (spec_decode_attn.py:184, self.part_o)
+  # measured here, VISION=1 VISION_OFFLOAD=0 SPEC=dflash2, RTX 3090 at 250 W. With the
+  # offload the same config comes up with the full 69,758-token pool and reads images.
+  #
+  # It is close to free: isolated-tower measurement at PCIe 4.0 x16, one 8192-patch image,
+  # median of 10 forwards, 891.3 -> 9.0 MiB of resident weights and 1160.5 -> 308.2 MiB of
+  # peak allocation for 296 -> 333 ms of encode, output bit-exact either way. Set
+  # VISION_OFFLOAD=0 only on a card with room to spare, where 36 ms per image buys nothing.
+  [ "${VISION_OFFLOAD:-1}" = 1 ] && export VLLM_VISION_CPU_OFFLOAD_GB=${VLLM_VISION_CPU_OFFLOAD_GB:-1}
 else
   VISION_ARGS="--language-model-only"
 fi
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -55,6 +55,29 @@ Things that each cost us hours, in rough order of pain. Worth skimming before yo
    between those two starts, which is the same run-to-run swing the V2 runner
    shows, so read the pool difference as noise rather than as a measurement of
    what the tower costs.
+
+   On `SPEC=dflash2` that 0.858 GiB is not optional headroom, it is the difference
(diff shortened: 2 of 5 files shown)


mhenrichsen:
Merged, with the verification run here first — and one precision added to the record in `1e497fe`, because re-testing your premise on this box turned up something worth keeping.

## Verified on a headless 3090 at 250 W, current main

`VISION=1 SPEC=dflash2 CTX=fast PREFIX_CACHE=1`, your patch applied:

| | measured |
|---|---|
| offloader engages | `Total CPU offloaded parameters: 0.08` (merger) + `0.85` (blocks) — your 870 MiB arithmetic, confirmed |
| KV pool | 68,605 tokens, **identical to `VISION=0`** on this tree (the 69,758 in the PR text is an earlier `KV_MEM` vintage; same "full pool" either way) |
| image round-trip | a marker word that exists only in the pixels comes back exact, 0.95 s |
| text decode | 120.6 tok/s at k=7 — no measurable cost, as you said |

API usage checked against the tree as well: `wrap_modules` mutates in place so the discarded return on the merger is fine, `uva_offloading` is a real attribute, and forcing it off per-instance leaves `--cpu-offload-gb` untouched exactly as your review note describes. The zero-copy trap note alone justifies the patch header's length.

## The one divergence: your no-boot repro needs the compositor

`VISION=1 VISION_OFFLOAD=0 SPEC=dflash2` — your OOM config — **boots and serves on this box.** Nothing else on the card here; your machine also carries Omarchy and a browser, and that share of the card is what pushes the tower over the margin. Before that reads as a weaker premise, look at where the resident-tower boot lands: **440 MiB free**. Gotcha 39's measured kill line for a concurrent burst is between 396 (dead) and 436 (survived) — so even headless, `VISION_OFFLOAD=0` is a config that boots, single-streams politely, and sits close enough to the cliff that I would not ship it. Your default-ON call is right in both geometries, for slightly different reasons, and gotcha 9 now records both.

Deploy note for anyone reading along: this is live on the reference box (launcher + patch), and the patch is inert at `VISION=0` — the shipped default is untouched.

Thanks for a contribution that came with its own negative result (the `SPEC=mtp` null, written as one) and the 11× zero-copy measurement. That is the house style.

Issue #33, opened 2026-08-23
Field report: OffloadingConnector (KV eviction to RAM/disk) works on this stack — 8 sharp edges and one open question


alexander99lab:
# Field report: OffloadingConnector (KV eviction to RAM/disk) on top of this stack — it works, with sharp edges

**Setup, so you can discount accordingly:** Ubuntu VM on Proxmox, GPU passthrough, RTX 3090 Ti, image built from `main-7db2d1f`, model = philbert440 AWQ body + our own GPTQ-int4 requant of lm_head/MTP (the "fast2" recipe from your drafter/README applied to a different body), bilingual ru-en draft vocab (18,195 ids), `SPEC=mtp DRAFT_TOKENS=3 PREFIX_CACHE=1`, bf16 KV. All numbers below are VM numbers; we already know from the PR #13 thread that this box pays 2x for PIECEWISE where bare metal pays <3%, so treat the speeds as internally comparable only.

**Why we tried it:** we want one process that keeps your speed but lets idle chats' KV spill to RAM/NVMe and come back without a full recompute (multi-user document-chat workload). Your 0.27.1 base already ships `vllm.v1.kv_offload` with `TieringOffloadingSpec`, an fs secondary tier, and — nice surprise — MTP-aware scheduling (`KV offloading: EAGLE/MTP draft attention groups [0..3] detected. The trailing chunk of these groups will be excluded from offloading due to volatility.`). So this is not a patch report; it is a "your image can already do this" report.

## What works

Boot flag (via `EXTRA_ARGS`):

```
--kv-transfer-config {"kv_buffer_size":268435456,"kv_connector":"OffloadingConnector","kv_role":"kv_both","kv_connector_extra_config":{"spec_name":"TieringOffloadingSpec","cpu_bytes_to_use":8589934592}}
```

plus `PYTORCH_CUDA_ALLOC_CONF=backend:native` (see edge 1). With that:

- MTP keeps working through offload cycles: acceptance 84–100%, mean accepted length 3.5–4.0, canary codes in evicted-and-returned documents all answered correctly.
- Return of an evicted ~10k-token document from the CPU tier: **1.1–1.3 s** vs **7–9 s** full recompute. Transfer rate ~6 GB/s (1.55 GB in 0.26 s).
- The fs secondary tier boots and cascades (`Created secondary tier #0 (fs)`) with `"secondary_tiers":[{"type":"fs","root_dir":...}]`.

[...]

PR #37, opened 2026-08-24
Run the uncensored checkpoint: fetch it, requantize it, and size dflash2 for it

danieldonaldson:
[philbert440/Qwen3.8-27B-Uncensored-Aggressive-W4A16-AWQ](https://huggingface.co/philbert440/Qwen3.8-27B-Uncensored-Aggressive-W4A16-AWQ) is an abliterated Qwen3.8-27B in the same `compressed-tensors` shape this repo already serves, with the vision tower and the grafted MTP head both preserved. `MODEL=` already points the launchers anywhere, so **serving** it needs no code — but **preparing** it does, and one of the two speculative modes needs a number.

### `prepare/quant_heads_stream.py`

Does what `quant_lm_head.py` + `quant_embed.py` + `quant_mtp.py` do, for a checkpoint those three cannot open. Two reasons, both properties of this export rather than of the model:

- **One 18.6 GB shard** (2384 tensors). The three scripts read a shard into a dict before rewriting it, which does not fit in RAM on this box. This streams the shard tensor-by-tensor and copies untouched tensors as raw bytes, so peak RSS is a few GB whatever the shard size.
- **Asymmetric AWQ body.** The three `deepcopy` `config_groups.group_0` and override only `num_bits`/`targets`, so the cloned group would declare `symmetric=false` / `zp_dtype=torch.int8` over the *symmetric* tensors they write — and vLLM would then look for a `weight_zero_point` that was never written. The groups written here always say `symmetric=true` / `zp_dtype=null`.

Same math and same output tensors otherwise. `bash verify.sh --no-server` with `MODEL=` set passes the result exactly as it passes the base model:

```
  PASS  lm_head requantized to int8      PASS  MTP draft module quantized
  PASS  embed_tokens requantized to int8 PASS  40k-token draft head present
```

### `prepare/fetch_uncensored.py`

Downloads it and prints the two commands that follow. It pulls `*.jinja` as well as `*.json`: `chat_template.jinja` is not `*.json`, and without it vLLM falls back to the tokenizer's built-in template, which does not emit the XML tool-call format `--tool-call-parser qwen3_coder` reads.

### `SPEC=dflash2` needs its pool resized

The one thing not discoverable from the model dir. After requantization this checkpoint is **15.68 GiB** of weights against the fast variant's **14.71**, and the DFlash2 branch pins the KV pool *in bytes* (`KV_MEM`) instead of sizing it from `--gpu-memory-utilization` — so the pool does not give that gigabyte back. The server loads, captures graphs, and dies on the split-KV verify buffer:

```
Model loading took 15.71 GiB
reserved 5.2 GiB memory for KV Cache as specified by kv_cache_memory_bytes config
torch.OutOfMemoryError: Tried to allocate 960.00 MiB ... 926.44 MiB is free
```

`KV_MEM=4456028569 DFLASH_MAX_LEN=98304` hands it back. Measured on one RTX 3090 at 250 W: a 4.15 GiB pool holding **103,033 tokens** at 98,304 `max-model-len` (4.6% margin), **85.7 tok/s** greedy on a 400-token answer, and images answered correctly with `VISION=1`. `SPEC=mtp` needs none of it — its pool is profiled, so it shrinks itself.

Docs only for the resize; no launcher behaviour changes, so nothing here can affect the base model or the fast variant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,65 @@ answer, and 400 tokens of fluent Danish inventing a task the prompt never asked 
 sweeps all 128 residues and judges every answer on how much of the document came
 back, and `bench/verbatim.py` self-tests that rule against all three shapes.
 
+### A different checkpoint: the uncensored build
+
+`MODEL=` points the launchers at any Qwen3.8-27B checkpoint in the same
+`compressed-tensors` shape. The one verified here is
+[philbert440/Qwen3.8-27B-Uncensored-Aggressive-W4A16-AWQ](https://huggingface.co/philbert440/Qwen3.8-27B-Uncensored-Aggressive-W4A16-AWQ)
+— an abliterated (de-refused) Qwen3.8-27B, W4A16 AWQ, with the vision tower and
+the grafted MTP head both preserved. Prepare it once, then serve it:
+
+```bash
+venv/bin/python prepare/fetch_uncensored.py          # ~18.6 GB
+venv/bin/python prepare/quant_heads_stream.py models/Qwen3.8-27B-Uncensored-W4A16
+venv/bin/python prepare/build_draft_vocab.py  models/Qwen3.8-27B-Uncensored-W4A16 \
+  --ids prepare/draft_vocab_ids.json
+
+MODEL=$PWD/models/Qwen3.8-27B-Uncensored-W4A16 SPEC=mtp CTX=long PREFIX_CACHE=1 \
+  MAX_LEN=100000 bash single-user/start_qwen.sh
+```
+
+It needs `prepare/quant_heads_stream.py` rather than the three `quant_*.py` steps
+the base model uses, for two reasons that are properties of the checkpoint and not
+of the model: it ships as **one 18.6 GB shard**, which the three scripts read into
+RAM whole before rewriting, and its body is **asymmetric AWQ**, which those scripts
+would copy onto the symmetric tensors they write — vLLM then looks for a
+`weight_zero_point` that was never written. The streaming script handles both and
+produces the same tensors otherwise; `bash verify.sh --no-server` with `MODEL=` set
+checks the result exactly as it checks the base model.
+
+**`SPEC=dflash2` needs its pool resized for this checkpoint.** After requantization
+it is 15.68 GiB of weights against the fast variant's 14.71, and the DFlash2 branch
+pins the KV pool *in bytes* (`KV_MEM`) rather than sizing it from
+`--gpu-memory-utilization`, so the pool does not give that gigabyte back. The server
+loads, captures graphs, and then dies on the split-KV verify buffer:
+
+```
+Model loading took 15.71 GiB
+reserved 5.2 GiB memory for KV Cache as specified by kv_cache_memory_bytes config
+torch.OutOfMemoryError: Tried to allocate 960.00 MiB ... 926.44 MiB is free
+```
+
+Hand that gigabyte back and it comes up. `CTX=long` (int8 KV) is the one to spend it
+on, because it buys roughly twice the context per byte of pool that `CTX=fast` does:
+
+```bash
+MODEL=$PWD/models/Qwen3.8-27B-Uncensored-W4A16 SPEC=dflash2 CTX=long PREFIX_CACHE=1 \
+  KV_MEM=4456028569 DFLASH_MAX_LEN=98304 bash single-user/start_qwen.sh
+```
+
+Measured here, RTX 3090 at 250 W: a 4.15 GiB pool holding **103,033 tokens** at
+98,304 `max-model-len` (4.6% margin) and **85.7 tok/s** greedy on a 400-token
+answer. The checkpoint keeps its vision tower, and that run had `VISION=1` — images
+came back described correctly — so the numbers are an upper bound on what the
+default `VISION=0` needs, which drops the tower's weights entirely.
+
+`SPEC=mtp` needs no `KV_MEM` of its own: its pool is profiled from `GPU_UTIL` rather
+than pinned, so it absorbs the extra gigabyte by shrinking the pool for you. It is the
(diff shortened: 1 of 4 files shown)


Dmtrii-tesla:
Hey! Is it possible to cut off mtp head of main model quant to free vram? I'm using dflash, so mtp is unnecessary

mhenrichsen:
Review in two parts, because this PR is two things: a requant tool the repo has needed for a while, and a decision about what the front README endorses — the first is yours and it is good, the second is mine and I owe you a straight answer on it shortly.

## The tool: `quant_heads_stream.py` holds up to a source read

Both of your premises about the existing scripts are accurate — `quant_lm_head.py` really does read the whole shard into a dict (`for k in f.keys(): tensors[k] = f.get_tensor(k)`) and really does `deepcopy` `config_groups.group_0` overriding only bits/targets, so an asymmetric AWQ body would poison the cloned group exactly as you describe. The streaming rewrite is careful where it needs to be: offsets re-planned contiguously and copied in original disk order, bf16 handled by byte-reinterpret rather than `.numpy()`, originals renamed (not copied) to `.bak-orig`, and the row-chunked `pack_to_int32` is safe to concatenate because packing is row-independent.

And the demand is not hypothetical: the reporter in #33 runs **exactly this checkpoint family** — a philbert440 AWQ body with a hand-rolled int4 requant of lm_head/MTP, i.e. they reimplemented this script's job privately because the repo's three couldn't open the export. That settles "does anyone need a generic path" for me.

Three notes at the level of nits, none blocking:

1. `emb_key = next(k for k in wm if k.endswith("embed_tokens.weight"))` — if a checkpoint ever carries a second match (an untied `mtp.embed_tokens`, say), `next` silently takes dict order. An `assert` that the match is unique would make the failure loud.
2. `qc["ignore"]` is filtered with `i != "lm_head"` — correct for this export's spelling, but `ignore` entries also appear as `re:.*lm_head` in the wild; worth either handling both or asserting the entry was actually removed.
3. `fetch_uncensored.py`'s `allow_patterns` has no `*.txt`, so a checkpoint that ships `merges.txt` without `tokenizer.json` would come down untokenizable. Works for this one; a one-line pattern addition future-proofs it.

Also noting what I have **not** done: run the 18.6 GB pipeline here. Your `verify.sh` transcript and the measured `KV_MEM`/pool numbers are exactly the right evidence, and I will reproduce the requant end-to-end before anything merges.

## The decision that is mine: what the README carries

The split I am weighing is: take `quant_heads_stream.py` (and the prepare/README section) framed as **"third-party checkpoints in the same shape"** — single-shard exports, asymmetric AWQ bodies, with yours as the worked, verified example — and decide separately whether the front README carries a named section and a dedicated fetch script for one particular third-party finetune. That second half is a question about what this repo endorses on its front page rather than about your code, which is why it gets a slower answer than the review did. If I land on the split: would you be willing to reframe the fetch script as `fetch_thirdparty.py <hf-repo>` with this checkpoint as its documented default example? Nothing else in the PR would need to move.

Either way — the `KV_MEM` arithmetic in your writeup (pinned pool does not give back a heavier body's gigabyte; `SPEC=mtp` profiles itself) is correct, matches gotcha 39's model of where that memory lives, and is exactly the note the next person serving any heavier body needs.

tomByrer:
@danieldonaldson Seems that that model includes vision, so that will eat up much context.

If you want uncensored, can you find another text-only model, or run Heretic on the regular model?

mhenrichsen:
Merged, with the split I proposed — and since you didn't get a chance to weigh in on the reframe, I did it on main rather than asking you to redo the branch: `fetch_uncensored.py` is now `fetch_thirdparty.py <hf-repo> [dst]` with your checkpoint as the documented default example, and the README section is "Third-party checkpoints", leading with the zero-prep option that appeared in the meantime (#45, leminkozey's ready-made build) and carrying your worked example + the `KV_MEM` arithmetic as the general path. Your commits are in as authored; the reframe is a follow-up commit. If any of it misrepresents what you built, say so and I'll fix it.

The promised end-to-end reproduction ran before the merge, on this repo's 3090 box, and it sharpened three things:

1. **`quant_heads_stream.py` works as advertised** — fetch, requant, `verify.sh --no-server` all green against the philbert440 export. One honesty correction to the PR text: peak RSS measured **9.7 GB** on the 18.6 GB shard (`/usr/bin/time -v`), not "a few GB". Still comfortably streaming (half the shard, flat in shard count), but a 16 GB host should not attempt it casually. The prepare/README now carries the measured number.
2. **Your own printed next-step crashed on your own layout** — `build_draft_vocab.py` assumed the base model's `model_extra_tensors.safetensors` exists, and a single-shard export processed by your script has no extras file, so the draft-head write died with FileNotFound. Fixed on main (the draft head becomes the file's first content). Your verify transcript showed the draft-head PASS, so I assume your local run had an extras file from an earlier experiment; fresh runs hit it deterministically.
3. **The serving recipe checks out**: your `KV_MEM=4456028569 DFLASH_MAX_LEN=98304` dflash2 boot came up healthy here and served — health 200 in 290 s with a cold compile cache for the new model path, 22.3 GiB resident (~1.3 GiB headroom, matching your 4.6% margin arithmetic), completions clean.

Also answering @Dmtrii-tesla's question from this thread — *"can the MTP head be cut from the main model quant to free VRAM when using dflash?"* — there is nothing to cut: the target architecture loads with `skip_prefixes=["mtp."]` (`qwen3_5.py`), so under `SPEC=dflash2` the grafted MTP tensors are never loaded to the GPU at all. They cost you ~850 MB of disk in the checkpoint, zero VRAM. (Under `SPEC=mtp` they load as the separate draft model, which is the point of keeping them.)

Thanks for a contribution that was honest about its own edges — that's what made the reproduction cheap to run.

PR #42, opened 2026-08-26
Fix the int4 KV Crash

lachhabw:
### Add int4-per-token-head support for Qwen3-27B on RTX 3090

The repository's vLLM 0.27.1 was patched for **int8** KV caching but not **int4**, so the single-user `int4` path crashed at startup. This commit adds the missing int4 compatibility patches so `--kv-cache-dtype int4_per_token_head` works end-to-end:

1. `attn_utils.py` — reshape int4 hybrid KV page to 848 (matches the head-dim halving that the Marlin int4 backend performs, where `head_size // 2` is real). Fixes:
   - `RuntimeError: shape '[367, 848, 8, 136]' is invalid for input of size 657288192`

2. `triton_unified_attention.py` — drop the `INT4_PER_TOKEN_HEAD only supports causal attention` assert so non-causal / DFlash2 draft attention can compile.

3. `int4_per_token_head.py` — thread `use_causal` / `use_per_seq_causal` into the int4 Triton kernel (previously int8-only).

4. `kv_cache_utils.py` — fix hybrid allocator to compute the int4 head size from `get_kv_cache_shape` (halved per-token-head), so the int4 page is sized correctly instead of mirroring int8.

### DIFF
--- a/patches/int4-fixes.patch
+++ b/patches/int4-fixes.patch
@@ -1,0 +1,301 @@
+--- a/v1/worker/gpu/attn_utils.py
++++ b/v1/worker/gpu/attn_utils.py
+@@ -259,7 +259,24 @@
+         )
+     else:
+         # No padding — safe to use a contiguous view.
+-        kv_cache = kv_raw_tensor.view(dtype).view(permuted_kv_cache_shape)
++        needed = prod(permuted_kv_cache_shape)
++        if (
++            kv_cache_spec.kv_quant_mode.is_per_token_head
++            and kv_raw_tensor.numel() > needed
++            and kv_raw_tensor.numel() % permuted_kv_cache_shape[0] == 0
++            and needed % permuted_kv_cache_shape[0] == 0
++        ):
++            nb = permuted_kv_cache_shape[0]
++            raw_per_blk = kv_raw_tensor.numel() // nb
++            want_per_blk = needed // nb
++            kv_cache = (
++                kv_raw_tensor.view(nb, raw_per_blk)
++                .narrow(1, 0, want_per_blk)
++                .reshape(permuted_kv_cache_shape)
++                .view(dtype)
++            )
++        else:
++            kv_cache = kv_raw_tensor.view(dtype).view(permuted_kv_cache_shape)
+ 
+     return kv_cache.permute(*inv_order)
+ 
+--- a/v1/attention/ops/triton_unified_attention.py
++++ b/v1/attention/ops/triton_unified_attention.py
+@@ -856,9 +856,6 @@
+     # Sub-byte packed mode (INT4) needs a bespoke kernel (split-dot +
+     # sub-byte unpack); everything else goes through the core kernel below.
+     if kv_quant_mode == KVQuantMode.INT4_PER_TOKEN_HEAD:
+-        assert use_causal and not use_per_seq_causal, (
+-            "INT4_PER_TOKEN_HEAD only supports causal attention"
+-        )
+         from vllm.v1.attention.ops.int4_per_token_head import (
+             unified_attention_int4,
+         )
+@@ -891,6 +888,8 @@
+             softmax_segm_output=softmax_segm_output,
+             softmax_segm_max=softmax_segm_max,
+             softmax_segm_expsum=softmax_segm_expsum,
++            use_causal=use_causal,
++            per_seq_causal_ptr=per_seq_causal_ptr,
+         )
+         return
+ 
+--- a/v1/attention/ops/int4_per_token_head.py
++++ b/v1/attention/ops/int4_per_token_head.py
+@@ -385,6 +385,9 @@
+     PACKING_FACTOR: tl.constexpr,
+     FP8_MIN: tl.constexpr = float8_info.min,
+     FP8_MAX: tl.constexpr = float8_info.max,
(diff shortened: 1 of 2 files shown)


mhenrichsen:
Verified, rebased, and landed as `04edd2b` — thank you, this is a real capability the repo did not have. You are credited as co-author on the commit. Closing rather than merging only because the branch as submitted cannot apply on a correctly-built venv; the mechanics below, so you can see exactly what survived and what I changed.

## What I reproduced

Your config on this repo's reference 3090 (shipped `prepare/` checkpoints — your launcher's model names are this repo's own, which made this easy):

- **Unpatched**: `RuntimeError: shape '[380, 848, 8, 136]' is invalid for input of size 680570880` at engine init — your `[367, 848, 8, 136]` with this card's block count. Same failure, same site.
- **Patched**: serves. **314,915 tokens of pool at `MAX_LEN=256000`** (1.23× concurrency) on one 24 GB card — for scale, `CTX=huge`/KVarN reaches 268k with custom kernels; this does more on stock machinery. Coherent prose, tool-calling round trip with well-formed arguments, lookup lane active on verbatim copy. Cost: ~20% decode vs the shipped FLASH_ATTN bf16 config on a short-prose probe (86 vs 104 tok/s e2e) — a fair capacity-for-speed trade, now documented in the README as an experimental profile.

## What changed in the rebase

Your `int4-fixes.patch` was cut against **stock** 0.27.1, so on a venv with this repo's patch stack applied, 3 of the 5 `kv_cache_utils.py` hunks fail and the other 2 would insert duplicate functions. Specifically:

1. **Kept, verbatim** (the real fixes): the `attn_utils.py` per-block narrow, the `triton_unified_attention.py` causal-only assert removal, and the `use_causal`/`per_seq_causal_ptr` plumbing through `int4_per_token_head.py`. These are now `patches/int4-kv-per-token-head.patch` with a header documenting the three mechanisms.
2. **Dropped — already shipped**: `_promote_indivisible_block_sizes` and `_prefer_padding_sliding_window_buckets` are this repo's `hybrid-sw-block-promote.patch` (the `(syv)` docstrings you carried over are from there). Your diff bundled them for a stock venv; here they'd double-define.
3. **Dropped — unreachable**: the mamba GCD-hash change fires only when `mamba_cache_mode != "align"` with prefix caching or a connector active, and both this repo's launchers and your own `alternative.sh` pass `--mamba-cache-mode align`. If you hit a real failure in a config that reaches that branch, please open an issue with the launch line — I'd rather ship it against a reproduction than speculatively.
4. **`alternative.sh` parameterized**: env overrides per house convention (`MODEL/DRAFT/PORT/GPU_UTIL/MAX_LEN/MAX_SEQS/DFLASH_TOKENS/LOOKUP/EXTRA_ARGS`), card selection via `CUDA_VISIBLE_DEVICES` instead of the hardcoded `GPU=1`, `GPU_UTIL` default 0.95 per your own comment (0.93 fails the 256k capacity check on a bare card here: 4.6 vs 4.87 GiB), the `DFLASH_TOKENS>7 → --no-async-scheduling` rule from the main launcher, and the stale-`/dev/shm` cleanup. I also dropped `VLLM_SPEC_DECODE_ATTN=1`: that patch documents itself bf16-KV-only and never engages on the TRITON+int4 path, so the line promised a fast path this profile doesn't have.

## What is still open, flagged in the README

Quality **at depth** on the int4 cache (needle at 100k+, GSM8K through long context) is unmeasured — KVarN has those numbers, this profile doesn't yet. If you run any of the quality harness on your box, that's the missing datapoint and I'll take it into the table.

One more practical note: your commit's git identity is `Your Name <teams@Lfada2i>` — you probably want `git config --global user.name/user.email` set before your next PR so your GitHub profile gets the credit directly.

lachhabw:
Thanks for the rebase and co-author credit.

**1. The dropped mamba hunk is reachable. Current main fails to boot without it.**

Reset to your latest commit and reinstalled vllm 0.27.1, shipped patches then run, it fails every boot:

```
AssertionError: Each KV cache group's real block_size must be divisible by hash_block_size.
block_sizes=[1696, 1696, 1696, 1696, 1696, 1696, 1696, 1696, 848], hash_block_size=1696
```

The guard tests `MambaSpec.block_size != cache_config.block_size`, not the mode string. With align, the global block is raised to 1696 and `hybrid-sw-block-promote` puts the drafter's sliding-window group at 848. Divergence under align, branch fires, hash=1696, 1696 does not divide 848, assert.

Fix, one line:

```python
-        return scheduler_block_size, scheduler_block_size
+        return scheduler_block_size, math.gcd(*group_block_sizes)
```

GCD=848 divides every group. With it the config boots correctly. Without it it never starts. Reset and re-applied several times, same result.

**2. Quality battery, tag `int4kv`, MAX_LEN=256000, GPU_UTIL=0.93:**

| metric | int4-KV | shipped reference |
|---|---|---|
| PPL en | 10.73 | 10.68–10.93 |
| PPL da | 10.87 | 10.85–11.29 |
| PPL code | 3.06 | 3.05–3.20 |
| PPL all | **8.075** | 8.044–8.34 |
| GSM8K (n=200, greedy) | **95.5%** | 95.0–96.5% |

Within noise of the shipped W4A16 rows.

PR #54, opened 2026-09-01
Number the patch files in the order they must apply

jnohlgard:
## Summary

`docker build` fails on a clean checkout. The apply loop is a plain `patches/*.patch` glob, so patches apply in filename order, but a few of the 24 are sequentially dependent: `prefill-attn-int8.patch`'s hunk context includes lines that `spec-decode-int8-kv.patch` adds. `p` sorts before `s`, so the glob reaches the dependent patch before its dependency and the hunk fails.

Three commits, 42 files, no hunk content touched:

1. **Number all 24 patches** `0001-…` through `0024-…` in their true apply order (pure renames, 0 insertions / 0 deletions). Filename order is then the apply order, so the loop, the Dockerfile, and `verify.sh` work unchanged.
2. **Point every reference at the new names** in 17 doc and script files and 19 patch preambles. Two references (`docs/long-context.md`, `single-user/start_qwen.sh`) named `spec-decode-attn-int8.patch`, a file that never existed; they now point at the real one, `0013-spec-decode-int8-kv.patch`.
3. **Document the convention where patching happens**: above the loop in `README.md` and in the `Dockerfile`, and in `docs/python-314.md`. A new patch takes the next free number; one that must apply earlier renumbers everything after it. 

## How the order was derived

The numbers follow the commits that introduced each file. `git log --diff-filter=A` over the 24 files gives a chronological sequence (2026-08-15 to 2026-09-01) that matches the numbering exactly, ties within a commit alphabetical. The stack grew by accretion: each patch was written against the ones already deployed, so introduction order is apply order.

## What is intentionally unchanged

- The old names inside the `+` lines of `0013` and `0015`. Those are comments in the applied source; `verify.sh` and `patches/_check_applied.py` match them against the deployed tree, which keeps the old names until the patches are regenerated.
- Historical point-in-time counts in the docs ("fifteen patches", "18 of 19").

## Verification

- `git diff` over the 24 patch files: changes only in preambles, never in hunk bodies.
- `bash -n` on the extracted `Dockerfile` RUN body; `python3 -m py_compile` on the changed Python.
- Grep for the 24 old names across the tree: the only survivors are the two in-hunk lines above.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,12 +25,15 @@ COPY docker/requirements.txt docker/requirements.txt
 RUN venv/bin/pip install -r docker/requirements.txt
 
 COPY . .
+# Patch apply order lives in patches/series: a few patches carry hunk context
+# that an earlier patch adds, so the glob order of patches/*.patch is wrong.
 RUN set -e; SP=$(venv/bin/python -c 'import vllm, os; print(os.path.dirname(vllm.__file__))' | tail -n1); \
-    for p in patches/*.patch; do \
-      case "$p" in \
-        patches/dflash2-backport.patch) echo "== skip $p (DFlash2 is native in vLLM 0.28.0)"; continue ;; \
+    sed -e 's/#.*//' -e 's/^[[:space:]]*//;s/[[:space:]]*$//' -e '/^$/d' patches/series | \
+    while IFS= read -r name; do \
+      case "$name" in \
+        dflash2-backport.patch) echo "== skip $name (DFlash2 is native in vLLM 0.28.0)"; continue ;; \
       esac; \
-      echo "== $p"; patch -p1 -d "$SP" < "$p"; \
+      echo "== $name"; patch -p1 -d "$SP" < "patches/$name"; \
     done; \
     bash kvarn/install.sh; \
     bash verify.sh --install
--- a/README.md
+++ b/README.md
@@ -780,11 +780,16 @@ venv/bin/python prepare/fetch_thirdparty.py
 venv/bin/python prepare/quant_heads_stream.py models/Qwen3.8-27B-Uncensored-W4A16
 
 # patch vllm (all compatible patches are written against 0.28.0; reapply after upgrades)
-for p in patches/*.patch; do
-  case "$p" in
-    patches/dflash2-backport.patch) echo "skip $p (DFlash2 is native in vLLM 0.28.0)"; continue ;;
+# Order is patches/series, one basename per line: a few patches carry hunk context
+# that an earlier patch adds, so the glob order of the directory is wrong. A new
+# independent patch goes on the last line; one that must apply before an existing
+# patch is listed before it.
+sed -e 's/#.*//' -e 's/^[[:space:]]*//;s/[[:space:]]*$//' -e '/^$/d' patches/series |
+while IFS= read -r name; do
+  case "$name" in
+    dflash2-backport.patch) echo "skip $name (DFlash2 is native in vLLM 0.28.0)"; continue ;;
   esac
-  patch -p1 -d venv/lib/python3.12/site-packages/vllm < "$p"
+  patch -p1 -d venv/lib/python3.12/site-packages/vllm < "patches/$name"
 done
 # optional: the KVarN 4/2-bit KV cache for 262k context (docs/long-context.md)
 bash kvarn/install.sh
--- a/docs/long-context.md
+++ b/docs/long-context.md
@@ -134,7 +134,7 @@ Two patches make that work, and neither changes anything at bf16:
   per head. The drafter's layers then keep a 16-token block while their page is padded to the
   full 1.71 MiB primary page: 385 blocks at 1.88% utilisation, a constant 5.2 GiB. The patch
   rounds those layers' block up (16 → 864) so their page covers the maximum instead.
-- **[spec-decode-attn-int8.patch](../patches/spec-decode-int8-kv.patch)** — the split-KV verify
+- **[spec-decode-int8-kv.patch](../patches/spec-decode-int8-kv.patch)** — the split-KV verify
   kernel reads the quantized cache and is wired into the Triton backend, which otherwise cannot
   split KV for a multi-query verify at all (`use_3d` is off whenever `max_seqlen_q > 1`, and
   every DFlash2 step is a verify). Per attention layer at 128k, 8 query tokens: 1.3 ms for this
--- a/docs/optimizations.md
+++ b/docs/optimizations.md
@@ -142,7 +142,7 @@ head_dim 256 has no faster sm86 alternative (FlashInfer measured within 1.5%,
 and it costs the split-KV verify path at decode).
 
 **int8-QK prefill attention** (`PREFILL_ATTN=int8`,
-`patches/triton-prefill-attn-int8.patch`) attacks what is left after the GEMMs: the
+`patches/prefill-attn-int8.patch`) attacks what is left after the GEMMs: the
 16 full-attention layers, whose head_dim of 256 pins FA2 at 54-57 TFLOPS on
 sm86 (85% of the card's practical fp16 mma rate — no fp16 rewrite can win).
 A Triton kernel runs QK^T on int8 tensor cores at 2x the fp16 rate,
--- a/docs/python-314.md
+++ b/docs/python-314.md
@@ -65,7 +65,10 @@ venv/bin/pip install 'vllm[bench]==0.27.1' huggingface_hub hf_transfer ninja \
   flashinfer-python flashinfer-cubin==0.6.13
 # model + requantization exactly as the README
 VP=$(venv/bin/python -c 'import vllm,os;print(os.path.dirname(vllm.__file__))')
-for p in patches/*.patch; do patch -p1 -d "$VP" < "$p"; done   # order matters
+sed -e 's/#.*//' -e 's/^[[:space:]]*//;s/[[:space:]]*$//' -e '/^$/d' patches/series |
+while IFS= read -r name; do
+  patch -p1 -d "$VP" < "patches/$name"
+done   # order: patches/series
 bash verify.sh --no-server
 ```
 
--- a/patches/check_vllm_series.sh
+++ b/patches/check_vllm_series.sh
@@ -5,10 +5,11 @@ set -euo pipefail
 #
 # Two passes, because two tools are in play and they answer different questions:
 #
-#   1. Every patch, in the glob order the Dockerfile and the README use, applied
-#      with GNU `patch` -- the tool that actually installs this stack. This is
-#      the pass that says "a clone of this repo still builds". It was missing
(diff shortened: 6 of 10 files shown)


mhenrichsen:
Thank you for this — you are attacking the right thing. `453104e` on main "fixed" this same build failure by renaming `prefill-attn-int8.patch` to `triton-prefill-attn-int8.patch`, purely so that `t` would sort after `s`. That is the accidental mechanism your PR replaces, and it is not a defensible way to encode a dependency. I want the ordering to be explicit. Two things before I can merge, one correction, and one design question.

## Correction: the dependency you name is not the one that broke

The PR says `prefill-attn-int8.patch`'s hunk context includes lines that `spec-decode-int8-kv.patch` adds. It is `spec-decode-attn.patch`. The trailing context of the failing hunk (`flash_attn.py`, `@@ -1038,6`) is:

```
                 # syv patch: split-KV Triton attention for speculative-decode
                 # batches (1 < queries/request <= 10). FA2 does not split the KV
                 # sequence when max_seqlen_q > 1, leaving most SMs idle.
```

which is `spec-decode-attn.patch`'s inserted comment. `spec-decode-int8-kv.patch` touches the same file, but what it adds there is the `_spec_attn_run` signature and the `k_scale_cache` arguments, none of which appear in that hunk's context. Your numbering satisfies the real constraint anyway (`0007-spec-decode-attn` before `0024-prefill-attn-int8`), so the ordering is right and only the stated reason is wrong — but please fix the reason, because the next person to move a patch will read it as the map.

## Blocking: the branch is three patches stale

You branched at `bb739e4`. Main is at `253c76a` and now carries **27** patches, not 24: `mamba-align-checkpoint-order.patch` (`dfbb550`), `spec-sampler-prewarm.patch` (`201d237`), and `prefill-attn-int8.patch` is already renamed to `triton-prefill-attn-int8.patch` (`453104e`). A rebase is needed, and the three new ones want numbers — the two new patches are independent of the stack and belong at the end; the renamed one is your `0024` under a name that should go back to `0024-prefill-attn-int8.patch` once the number carries the ordering.

Once it is rebased I will verify the full set applies to the stock 0.27.1 wheel in the new order and merge on green.

## Design question: numbers, or a series file?

Your own PR names the cost: "one that must apply earlier renumbers everything after it." That is not hypothetical — `453104e` inserted a patch mid-stack last week, and under this scheme it would have renamed every file after it plus every reference in 17 docs. The churn falls on exactly the change that is already delicate.

The alternative is quilt's: keep the descriptive names, add `patches/series` listing them in apply order, one per line, and have the Dockerfile and `verify.sh` read that file instead of globbing. Inserting mid-stack edits one line. Renames stay rare, so doc references stay valid. The ordering is still explicit, still reviewable in a diff, and additionally it becomes checkable — a patch present in `patches/` but absent from `series` is a build error, which the glob cannot express and numbering cannot either.

I lean to `series` for that reason, but I recognise you have written the numbering version already and it does solve the problem. Your call; I will take either. If you keep the numbers, please also make the convention enforceable rather than documented — a check that the prefixes are dense and gapless and cover every `.patch` in the directory, wired into `verify.sh`, so a patch added without a number fails loudly instead of sorting wherever its name lands.

## Keep

Good catch on `docs/long-context.md` and `single-user/start_qwen.sh` pointing at `spec-decode-attn-int8.patch`, a file that has never existed. That is worth landing whichever shape the rest takes. Likewise the "changes only in preambles, never in hunk bodies" discipline, and verifying it with a diff rather than asserting it.

mhenrichsen:
Update, because something happened this week that makes your case stronger than my counter-proposal did.

**The patch-integrity job was checking five of thirty patches, and applying none of them.** `git apply` resolves a patch path against the *repository* root and silently skips anything outside the subdirectory it runs in; the job ran `git -C .ci/vllm/vllm apply` while the patches address files relative to the installed vllm package, so every path fell outside the prefix, every file was skipped, and the script printed `patch integrity: OK` having changed nothing. Fixed in `d21ed22`, and extended in `7ca84fc` to apply **all thirty in glob order with GNU patch** against a pristine v0.28.0 checkout. Result: 30 applied, 0 failed. Both passes now assert the tree actually changed, so the vacuous case cannot come back.

Why that matters here: the ordering you are trying to make explicit is currently encoded in **three** places that must agree and are not checked against each other — the Dockerfile's `for p in patches/*.patch`, `docs/python-314.md`'s copy of the same loop with the comment `# order matters`, and a hardcoded five-element array in `check_vllm_series.sh`. `453104e` renaming `prefill-attn-int8.patch` to `triton-prefill-attn-int8.patch` so that `t` sorts after `s` is the fourth. A `patches/series` file collapses all four into one, and it now has a CI job that would actually notice if it were wrong.

**So I would like to land this, in the series-file shape rather than the numeric-prefix shape**, for the reason I gave in September: renumbering rewrites filenames that appear in issue threads, commit messages, gotchas and `verify.sh` output, and a dependency between two patches is not a total order over thirty of them.

The branch has gone stale — `CONFLICTING` against main now. Two ways forward, your call:

1. You rebase and swap the numeric prefixes for `patches/series`, and I merge it.
2. You would rather not, and I implement it and credit you in the commit. The analysis is yours; `453104e` was mine and you are the one who caught what it really was.

Say which and I will act on it. If I do not hear back I will take option 2 rather than leave the ordering implicit, but I would much rather it were option 1.

One correction from my earlier comment still stands and is worth carrying into whichever version lands: the dependency is `spec-decode-attn.patch`, not `spec-decode-int8-kv.patch`.

jnohlgard:
My coding agent is working on a rebase round and will incorporate your suggestions in the revised branch. I will post back when I have checked the results.

jnohlgard:
@mhenrichsen Option 1, done: the branch is rebased and in the `patches/series` shape now. Two commits on `1834917` (current main).

**`patches: apply in the order of patches/series, not glob order`**

- New `patches/series`, one basename per line in apply order. A mid-stack insertion edits one line; nothing gets renamed.
- Every site that used to encode the order now reads the file: the Dockerfile loop, the README install loop, the `docs/python-314.md` loop, and `verify.sh`. `check_vllm_series.sh` pass 1 applies in series order, and its pass 2 contractual DFlash order is derived from the series (fixed set, order taken from the file) so the two can't drift.
- Make it checkable as you asked: both `verify.sh` and the CI script fail if `series` and `patches/` disagree in either direction. A patch added without a series entry now fails loudly at install/verify time instead of sorting wherever its name lands.
- `triton-prefill-attn-int8.patch` goes back to `prefill-attn-int8.patch` now that the rename only existed to fix the sort order. Its preamble keeps the dependency attribution to `spec-decode-attn.patch` (the real one, per your correction) and records why the triton- name existed.
- In-hunk references to the old names stay untouched, as before: they match the deployed tree, which `verify.sh` and `_check_applied.py` check against.
- The series reader is POSIX throughout (`sed` for comments/whitespace, `while IFS= read -r`); the Dockerfile RUN body passes `sh -n` and `bash -n`.

**`docs: point at the real spec-decode-int8-kv.patch, not the name that never existed`**

- `docs/long-context.md` and `single-user/start_qwen.sh` referenced `spec-decode-attn-int8.patch`, which never existed; both now point at `spec-decode-int8-kv.patch`.

On the moved base:

- #92 (the `.orig` hunk removal) landed first, so this PR no longer contains that work; I simply rebased on top of it.
- #89 added `vllm-pr54282-draft-gumbel-salt.patch`; it is now the last entry of the series. Its preamble expects `dflash2-lookup-drafting.patch` applied before it, which the tail position satisfies.
- #97 (hf xet variable) came in with the rebase; no conflicts.

Verification, against a pristine vLLM 0.28.0 clone:

- `check_vllm_series.sh`: pass 1 applies 31/31 (backport skipped), pass 2 green, `patch integrity: OK`.
- The README loop, copied verbatim from the doc, rebuilds the patched tree (rc 0) — the new gumbel-salt patch is picked up through the series file without the loop changing at all.
- `verify.sh` patch section: 34/34 PASS, including the new series/directory consistency check. Negative test: dropping one line from `series` fails the check and prints the diff.

mhenrichsen:
Merging. This is the shape I asked for, and the ordering claim is now proved by CI rather than asserted.

What I checked, on a fresh clone with the PR head (7ff5ff3) on top of current main (1834917):

- `grep -v '^\s*#' patches/series | grep -v '^\s*$' | wc -l` -> 32, `ls patches/*.patch | wc -l` -> 32, and `comm -3` between the two sorted lists is empty, so the new bidirectional check passes on its own tree.
- The dependency I flagged is satisfied: entry 7 is `spec-decode-attn.patch`, entry 24 is `prefill-attn-int8.patch`.
- CI run 34597805233 (git-apply) is green: `check_vllm_series.sh` pass 1 applies 31/31 in series order with GNU patch against a pristine vLLM v0.28.0 checkout (backport skipped), pass 2 clean, "patch integrity: OK". That is the job that was itself fixed in d21ed22/7ca84fc, so it is actually exercising the paths now.
- No stale names survive: `grep -rI 'triton-prefill-attn-int8\|spec-decode-attn-int8'` returns nothing. The only remaining `patches/*.patch` globs are verify.sh:56 and patches/check_vllm_series.sh:46, which are the ON_DISK side of the consistency check, plus one Dockerfile comment.
- `bash -n verify.sh`, `bash -n patches/check_vllm_series.sh`, and `sh -n`/`bash -n` on the extracted Dockerfile RUN body are all clean. verify.sh cds to `$HERE` at line 14, so the relative `patches/series` read is safe.

Credit where it is due: you found a real latent breakage (glob order reaching a dependent patch before its dependency), you derived the order from `git log --diff-filter=A` rather than guessing, and when I pushed back on the numeric-prefix scheme you rewrote it to the series file instead of arguing. The rename back to `prefill-attn-int8.patch` and the two references to a file that never existed are both correct fixes.

What is not proved, and what I am not asking you to prove: this is packaging and ordering only, no hunk body, no kernel, no default flag. There is no throughput or acceptance measurement that this diff could move, so I did not run one. docs/MR-DRAFT.md's owed checks are about the VLLM_INT4_MQ_3D default flip and are untouched here - nothing discharged, nothing newly owed.

Two things I am handling on my side, noted so nobody is surprised:

1. The title and body still describe the old "number all 24 patches 0001-...0024-" scheme that the branch no longer implements. The two commit messages are accurate, so I am merging on those, not on the body. No action needed from you.
2. Merge-order coupling: #101 and #90 each add a patch file. Once this lands, a patch added without a series line fails verify.sh and CI by design. I will tell both authors to append one line (tail position is right for both; #101's preamble says it applies on top of mamba-align-checkpoint-order.patch, which is line 26).

And a deploy note for me, not for this PR: the deploy tree is not a git repo and syncs with plain `rsync -a` (no `--delete`), so after this merge it will hold both the old and new prefill patch filenames and the ON_DISK check will fail there on an extra file. Fix is one targeted `rm` of the stale file after the sync, not an `rsync --delete`. I will add that to docs/gotchas.md, which currently says nothing about apply order at all.

Issue #48, opened 2026-08-29
Engine core stalls on the first large chunked-prefill after boot (WSL2, KVarN + DFlash2)


ApatheticMioz:
## Platform

**WSL2 (Ubuntu, kernel 6.18.33.2-microsoft-standard-WSL2, NVIDIA driver 616.56,
single RTX 3090 24GB).** Everything in this report was observed on WSL2, and our
leading suspicion involves the WSL2 GPU stack, so the platform is a first-class
part of the bug report, not a footnote. If this cannot be reproduced on bare
metal, WSL2 itself is a primary suspect.

## Summary

With this repo's `SPEC=dflash2 CTX=huge PREFIX_CACHE=1` single-user config
(KVarN 4/2-bit KV, DFlash2 7-token speculative decoding; vLLM 0.27.1 at commit
2ae239f), the vLLM engine core can **wedge on the first large chunked-prefill
request after a fresh boot**:

- the unconditional 10-second `Engine 000: … Running:` stats lines go silent;
- the API process stays up — the port answers, requests are accepted, but none
  is ever scheduled;
- in one production incident (2026-08-28) this persisted **~4.5h** with the GPU
  at 100% (~24.1/24 GB resident) producing nothing, until the core spontaneously
  unwedged; the operator then stopped the engine.
- over the following ~36h the same failure recurred **5 times**, each minutes
  after a fresh boot, always during a large-prompt (10k+ token) request.

A one-request reproduction harness, an A/B across three configurations, and a
py-spy capture of the wedged core are below. We did not root-cause it to a
single line, so this reports the phenomenon + evidence rather than a
speculative fix, and ends with two questions for the maintainer.

## Environment

- WSL2 (Ubuntu), kernel 6.18.33.2-microsoft-standard-WSL2, NVIDIA driver 616.56,
  RTX 3090 24GB, 250 W
- vLLM 0.27.1 + this repo's patches, at commit **2ae239f** (pristine tree; see
  "local patch" note below)
- Launch: the repo's `SPEC=dflash2 CTX=huge PREFIX_CACHE=1`
  `single-user/start_qwen.sh` config, i.e.
  `vllm serve <Qwen3.8-27B> --served-model-name qwen3.8-27b
  --gpu-memory-utilization 0.93 --max-model-len 245760 --max-num-seqs 8
  --kv-cache-dtype kvarn_k4v2_g128 --block-size 128
[...]

PR #55, opened 2026-09-02
single-user: add optional serving warmup after health check (WARMUP)

lilidl-nft:
Adds an optional single-user serving wrapper that boots the launcher, waits for /health, runs a controlled serving-path warmup (bench/warmup.sh), and only then considers the server ready for production traffic.

The launcher ends in 'exec vllm serve', so a systemd unit has no after-boot hook. A cold boot may still JIT-compile Triton / speculative-decoding variants on the first serving requests (the repo's #48 prewarm covers most; the rest still compile in request 1). On tightly packed 24 GB cards those transient allocations can OOM the engine even after /health is 200 (gotcha 39's 'boots, /health 200, dies on the first prompt' shape). This complements -- not replaces -- the #48 kernel prewarm.

- bench/warmup.sh: a small decode pass (2x128, C1), then a concurrent one (8x256, C4); exit 1 if the server is not healthy or a pass fails.
- single-user/qwen-server.sh: WARMUP=1 (default) does the health wait + warmup; WARMUP=0 restores the legacy boot. The wrapper forwards SIGTERM/INT and exits with the server's code, so Restart=on-failure is unchanged. WARMUP_ATTEMPTS/INTERVAL tune the health-wait window.
- qwen-serving.service + README: unit opts in (Environment=WARMUP=1); a 'Post-boot serving warmup' section documents it.

start_qwen.sh is left untouched.

### DIFF
--- a/bench/warmup.sh
+++ b/bench/warmup.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+# Post-boot serving warmup: exercise the serving path (decode, speculative
+# verify, small and concurrent batches) before production traffic.
+#
+# The launcher's boot-time profile runs only profiling dummies, so a cold
+# Triton cache may still compile serving-path variants on the first real
+# request (gotchas 8, 39, 50). This script spends that cost in a controlled
+# pass: one small request first (decode + speculative-verify capture), then a
+# larger concurrent batch (the multi-sequence verify path).
+#
+# Exits 1 if the server is not healthy or any pass fails, so the serving
+# wrapper (single-user/qwen-server.sh) treats it as a failed boot.
+#
+#   bash bench/warmup.sh             # 127.0.0.1:18020, the launcher's model
+#   PORT=18021 bash bench/warmup.sh  # a different port
+set -euo pipefail
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO="$(dirname "$HERE")"
+cd "$REPO"
+export PATH="$REPO/venv/bin:$PATH"
+
+# Reuse the API key the launcher started the server with. start_qwen.sh
+# exports VLLM_API_KEY (from $VLLM_API_KEY or api_key.txt), and vLLM reads
+# that variable to bind --api-key; vllm bench serve presents OPENAI_API_KEY,
+# so the two must agree or every request 401s. An explicit OPENAI_API_KEY
+# wins, then VLLM_API_KEY, then api_key.txt, then a harmless placeholder (the
+# server ignores the key when none was bound).
+if [ -z "${OPENAI_API_KEY:-}" ]; then
+    if [ -n "${VLLM_API_KEY:-}" ]; then
+        export OPENAI_API_KEY="$VLLM_API_KEY"
+    elif [ -f "$REPO/api_key.txt" ]; then
+        export OPENAI_API_KEY="$(cat "$REPO/api_key.txt")"
+    else
+        export OPENAI_API_KEY="EMPTY"
+    fi
+fi
+
+HOST=${HOST:-127.0.0.1}
+PORT=${PORT:-18020}
+# Resolve the model the way single-user/start_qwen.sh does: the fast variant
+# when present, else the base dir. Only the tokenizer matters here (its vocab
+# size drives random-token generation; no weights are loaded), and the request
+# is matched on --served-model-name, so any variant of the same model works.
+MODEL=${MODEL:-}
+if [ -z "$MODEL" ]; then
+    MODEL="$REPO/models/Qwen3.8-27B-W4A16-AutoRound"
+    if [ -d "$REPO/models/Qwen3.8-27B-W4A16-AutoRound-fast" ]; then
+        MODEL="$REPO/models/Qwen3.8-27B-W4A16-AutoRound-fast"
+    fi
+fi
+
+# Build the command as an array, not a string: an unquoted "$B" re-splits and
+# re-globs, so a path with a space or glob character would break, and a
+# quoted "$B" would not word-split at all. "${BENCH[@]}" expands each element
+# verbatim. (venv/bin/vllm is relative because we cd "$REPO" above.)
+BENCH=(venv/bin/vllm bench serve --host "$HOST" --port "$PORT" --model "$MODEL" --served-model-name qwen3.8-27b)
+
+curl -sf -o /dev/null "http://$HOST:$PORT/health" || { echo "[warmup] no server on $HOST:$PORT" >&2; exit 1; }
+echo "[warmup] server ready, warming the serving path"
+"${BENCH[@]}" --dataset-name random --random-input-len 128 --random-output-len 128 --num-prompts 2 --max-concurrency 1 --ignore-eos --temperature 0 > /dev/null 2>&1 \
+  || { echo "[warmup] small-batch pass failed" >&2; exit 1; }
+"${BENCH[@]}" --dataset-name random --random-input-len 256 --random-output-len 256 --num-prompts 8 --max-concurrency 4 --ignore-eos --temperature 0 > /dev/null 2>&1 \
+  || { echo "[warmup] concurrent-batch pass failed" >&2; exit 1; }
+echo "[warmup] done"
--- a/single-user/README.md
+++ b/single-user/README.md
@@ -369,6 +369,29 @@ systemctl --user enable --now qwen-serving
 loginctl enable-linger $USER
 ```
 
+The unit runs `qwen-server.sh`, which boots the launcher, waits for
+`/health`, runs `bench/warmup.sh` (the optional post-boot serving warmup,
+`WARMUP=1` is the unit's default; `WARMUP=0` restores the legacy boot
+behavior), and only then considers the server ready for traffic.
+
+### Post-boot serving warmup
+
+For tightly packed 24 GB deployments, `/health` may become available before
+all serving-path Triton / speculative-decoding variants have been exercised
+(the launcher's boot-time profile runs profiling dummies only). The optional
+wrapper waits for health, runs `bench/warmup.sh` (a small decode pass, then a
+concurrent one), and only then considers the server ready for traffic — it
+complements the kernel prewarm in #48 and does not replace it:
+
+```bash
+WARMUP=1 bash single-user/qwen-server.sh   # wait /health, warm, then serve
+WARMUP=0 bash single-user/qwen-server.sh   # legacy boot, no warmup
+```
+
+The included systemd service enables this by default
+(`Environment=WARMUP=1`). `WARMUP_ATTEMPTS` / `WARMUP_INTERVAL` tune the
+health-wait window (30 × 5 s by default).
+
 Point your chat client at `http://<host>:18020/v1` with the key from
 `api_key.txt`. Works with anything that speaks the OpenAI API, tool calling
 included (`tools` + `tool_choice: "auto"` come back as `tool_calls`).
@@ -392,7 +415,8 @@ included (`tools` + `tool_choice: "auto"` come back as `tool_calls`).
 | `TOOLS` | 1 | tool/function calling (`--enable-auto-tool-choice --tool-call-parser`). `TOOL_PARSER` (`qwen3_coder`) must match the XML call format this model's chat template emits — `hermes` parses the JSON a Qwen model does *not* produce here, and fails silently. 0 = off, and `tool_choice: "auto"` then 400s |
 | `VISION` | 0 | 1 keeps the vision tower instead of `--language-model-only` (0.858 GiB of BF16 weights on this checkpoint), for a client that sends images: one image per prompt and a 2048-image-token pixel cap, both overridable from `EXTRA_ARGS` |
 | `VISION_OFFLOAD` | 1 | with `VISION=1`, keeps the tower's weights in pinned host RAM and copies each module to the GPU for its own forward (`patches/vision-tower-cpu-offload.patch`). **On 24 GB, `SPEC=dflash2` + `VISION=1` does not boot with this off** — the tower is 0.85 GiB of the ~1.1 GiB transient margin, and graph capture OOMs allocating the 960 MiB split-KV verify buffer with 787 MiB free. With it on, the same config comes up at the full 69,758-token pool and reads images. Costs 296 → 333 ms of encode per 8192-patch image, output bit-exact. 0 only on a card with headroom to spare. `VLLM_VISION_CPU_OFFLOAD_GB` (default 1) is the budget in GiB |
-| `REQ_METRICS` | 0 | 1 = `--enable-per-request-metrics --enable-force-include-usage`: per-request timing fields in every response and `usage` on every request, the fields llama-swap's dashboard reads ([#51](https://github.com/syv-ai/qwen38-27b-rtx3090/issues/51)). `prompt_tokens_details.cached_tokens` is always on. Not compatible with `--disable-log-stats` in `EXTRA_ARGS`; vLLM's per-request spec-decode summary flag is nightly-only, not in 0.27.1 |
+| `REQ_METRICS` | 0 | 1 = `--enable-per-request-metrics --enable-force-include-usage`: per-request timing fields in every response and `usage` on every request, the fields llama-swap's dashboard reads ([#51](https://github.com/syv-ai/qwen38-27b-rtx3090/issues/51)). `prompt_tokens_details.cached_tokens` is always on. Not compatible with `--disable-log-stats` in `EXTRA_ARGS`; vLLM's per-request *spec-decode* summary flag is nightly-only, not in 0.27.1 |
+| `WARMUP` | 1 | wrapper-only: 1 = wait `/health` then run `bench/warmup.sh` before serving (see "Post-boot serving warmup"); 0 = legacy boot. The unit sets this explicitly. Not read by `start_qwen.sh` itself |
 | `PORT` | 18020 | |
 
 ## Switching modes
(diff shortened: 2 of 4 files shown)


mhenrichsen:
I ran this on the reference 3090. It works end to end — boot, health, warmup, `server ready for traffic`, serving, 55s total — and `bench/warmup.sh` is carefully written (the `BENCH=(...)` array with the comment about why, the `OPENAI_API_KEY` precedence chain, re-deriving the `-fast` model the way the launcher does). I am not merging yet: one defect makes the wrapper unsafe as a systemd `ExecStart`, and the stated motivation does not reproduce.

## Blocking: SIGTERM kills the wrapper and leaves the server running

Sending `SIGTERM` to the wrapper — which is exactly what `systemctl stop` does — gives, 25 s later:

```
after SIGTERM: wrapper alive=no; server alive=yes
```

The wrapper exits and the vLLM process survives, holding all 23 GB of VRAM. That is worse than the problem being solved, because `qwen-serving.service` has an `ExecStartPre` that waits for `memory.used < 1000 MiB` and fails after 120 s: the orphan blocks its own unit from restarting, so `Restart=on-failure` cannot recover. Today `ExecStart` is the server itself, so systemd kills it directly and this cannot happen.

The cause is the interaction between the trap and `wait`. The script is blocked in `wait "$SERVER_PID"` when the signal arrives; bash runs `cleanup`, whose own `wait` does not block as intended on a child whose `wait` was already interrupted, so the script falls out and exits while the server is still shutting down. `KillMode=mixed` in the unit would paper over it, but the wrapper should not depend on that. The robust shape is to loop until the child is actually gone:

```bash
cleanup() {
    kill -TERM "$SERVER_PID" 2>/dev/null || true
    for _ in $(seq 1 60); do kill -0 "$SERVER_PID" 2>/dev/null || return; sleep 1; done
    kill -KILL "$SERVER_PID" 2>/dev/null || true
}
```

and to add `EXIT` to the trap so no path can leave an orphan.

## The premise does not reproduce: zero in-request JIT, warm cache or cold

The PR says the #48 prewarm "covers most of them; the ones it misses still compile in request 1". I measured that with #48's own instrument — `--jit-monitor-verbose`, counting `jit_monitor` lines at `/health` and again after request 1 — on the production profile (`start_qwen.sh`, KVarN, `CTX=fast`, `DFLASH_TOKENS=15`, prefix caching), against a 17,871-token prompt. Cold arms used throwaway `TRITON_CACHE_DIR` and `VLLM_CACHE_ROOT`, so nothing was reused:

| arm | boot to health | JIT during request 1 | JIT during warmup |
|---|---:|---:|---:|
| warm cache, no warmup | 35 s | **0** | — |
| warm cache, warmup | 30 s | **0** | 0 |
| cold cache, no warmup | 125 s | **0** | — |
| cold cache, warmup | 126 s | **0** | 0 |
| warm cache, warmup (rerun) | 55 s | **0** | 0 |
| cold cache, warmup (rerun) | 149 s | **0** | 0 |

Zero in six arms, and the warmup itself compiles nothing. An empty compile cache costs about 90 s **at boot**, before `/health` — it does not leak into request 1. So on this profile there is nothing left for a post-boot warmup to prewarm; #48 closed it. (In my first pass both warmup arms failed with `small-batch pass failed` — that was my harness running your scripts out of a scratch directory, so `REPO` resolved to a tree with no venv. My error, not yours; the rerun above is from a proper checkout.)

**What survives, and I think it is the real case for this PR:** your second motivation is memory, not JIT — "those transient allocations can OOM the engine even after /health is 200". That is gotcha 39's shape and it is a genuine failure mode this repo has hit. A warmup converts a first-user-request OOM into a boot-time failure that systemd can act on, which is worth having even at zero JIT. I would like the PR to lead with that, since it is the part that holds, and drop or heavily qualify the JIT claim. If you have a profile where you *measured* in-request compiles, please post the counts — I tested the KVarN path and would want to fix any profile where it still happens, in the prewarm rather than here.

## Design: a warmup failure should not take down a healthy server

Both the health timeout and a failed warmup do `cleanup; exit 1`, killing an engine that is serving fine. With `Restart=on-failure` a transient bench failure becomes a restart loop, and the health timeout is the sharper edge:

- `WARMUP_ATTEMPTS=30 × WARMUP_INTERVAL=5` gives a 150 s window. My cold-cache boots took **125 s and 126 s**, and one took **149 s**. That is a margin of a second, and it kills the server on expiry. A first boot after a host reboot, with the page cache cold on a 27B checkpoint, will exceed it.

Warmup is an optimisation, so its failure should be loud but not fatal: log `[server] warmup failed, serving anyway` and fall through to `wait`. For the health wait, either drop the timeout entirely (systemd's `TimeoutStartSec` already bounds a boot that never completes) or raise the default well past the slowest real boot — 600 s would be defensible — and on expiry log and continue rather than kill.

## Smaller notes

- `trap cleanup INT TERM` omits `EXIT`; with `set -e` any unexpected failure between the launch and the final `wait` orphans the server. Add `EXIT`.
- `warmup.sh` re-derives `MODEL` and the API key by duplicating `start_qwen.sh`'s logic. It is correct today and the comments say why, but it will drift. Sourcing a small shared resolver, or having the wrapper export what it already knows, would keep them honest.
- The two warmup passes cost 24–25 s. Worth stating in the README next to `WARMUP=1`, since it is added to every boot.

Fix the SIGTERM handling and the fatal-failure behaviour, re-aim the motivation at the OOM case, and I will re-run this and merge. Thanks for the wrapper — the `exec`-means-no-post-boot-hook observation is the right problem to have noticed.

mhenrichsen:
Correction to my comment above, on the two rerun rows and one conclusion I drew from them.

The last two rows of the JIT table report **boot plus warmup**, not boot to health. My second-pass harness logged a single timestamp at the end of the warmup instead of one at `/health`, and I read it as a boot time. Corrected:

| arm | boot to health | JIT during request 1 |
|---|---:|---:|
| warm cache, warmup (rerun) | ~30 s (55 s including the 25 s warmup) | 0 |
| cold cache, warmup (rerun) | ~125 s (149 s including the 24 s warmup) | 0 |

So **no boot took 149 s**, and "that is a margin of a second" is wrong. The four directly measured boot-to-health times are 35, 30, 125 and 126 s, and the two rerun boots agree with them. Against the 150 s default window the real margin on a cold compile cache is about 24 s, not one second.

The recommendation does not change, but the reason is now the weaker one I should have given: a 24 s margin on a 125 s boot is thin, it is measured on a *warm page cache* with the checkpoint already resident, and the consequence of crossing it is that the wrapper kills a server that would have become healthy. A first boot after a host reboot has to read a 27B checkpoint off disk on top of the 125 s. Raise the default and make expiry non-fatal.

The JIT counts themselves are unaffected — zero in all six arms, which is what the table is for. So is the SIGTERM orphan, which was a separate observation.

mhenrichsen:
Status check, five days on, with the disposition spelled out so this does not sit here indefinitely.

Where it stands, from my testing above: the wrapper works end to end and `bench/warmup.sh` is carefully written. Two things stop it:

1. **Blocking, and fixable in three lines.** `SIGTERM` to the wrapper leaves the vLLM process running and holding 23 GB (`wrapper alive=no; server alive=yes`). That makes it unsafe as a systemd `ExecStart`, because `qwen-serving.service` has an `ExecStartPre` that waits for `memory.used < 1000 MiB` and would then block the next boot for its full timeout. The fix is a trap that forwards the signal and waits:

   ```bash
   trap 'kill -TERM "$SERVER_PID" 2>/dev/null; wait "$SERVER_PID"' TERM INT
   ```

   plus `wait "$SERVER_PID"` as the last line instead of falling off the end. Worth checking with `systemctl --user stop` rather than a bare `kill`, since systemd sends to the whole control group and the failure mode differs.

2. **The stated motivation does not reproduce, and I corrected my own numbers on that.** The four measured boot-to-health times were 35, 30, 125 and 126 s against the 150 s default window, so the real margin on a cold compile cache is about 24 s — not the one second I first wrote. Zero JIT events during request 1 in every arm, warm or cold, with or without the warmup. So the warmup is not preventing JIT; #48 closed at zero JIT and `spec-sampler-prewarm.patch` is what handles that.

That leaves a real but narrow benefit: about 24 s of margin against the health-check window on a cold cache, which matters on a first boot in a fresh container and not otherwise. That is worth having as an **opt-in** `WARMUP=1`, which is what you built, provided it cannot orphan a server.

**So**: fix the trap and I will merge it. If you would rather not, say so and I will close this with the measurements written into the docs as the cold-cache boot budget, which is the durable part of what you found either way — that a first boot in a fresh container comes within 24 s of the default window is worth someone knowing before they hit it.

No rush on my side, but I would like to know which way you want it to go.

lilidl-nft:
Hi! I will fix it. Give me this week and I'll send it.

mhenrichsen:
Still open, still waiting on you — and I owe you an update on one thing, because the repo has since answered a question I asked you and the answer partly goes your way.

Where the branch is: one commit, `650d6ef`, `2026-09-02`. Nothing has landed since your "give me this week" on Sep 7, so the two blocking defects from my Sep 2 review are still literally in the diff:

```
$ gh pr view 55 --json headRefName,commits,updatedAt
fix/warmup-after-boot  650d6ef2 (2026-09-02T00:48Z)  updated 2026-09-07
```

**1. The orphan.** `single-user/qwen-server.sh` still has `trap cleanup INT TERM` with no `EXIT`, and `cleanup()` is `kill -TERM "$SERVER_PID"` followed by `wait "$SERVER_PID" 2>/dev/null || true`. That `wait` does not block — the foreground `wait` was already interrupted by the signal — so the wrapper returns while the engine is still shutting down, which is how I got `wrapper alive=no; server alive=yes` holding 23 GB. Forward the signal and actually block:

```bash
cleanup() {
    kill -TERM "$SERVER_PID" 2>/dev/null || true
    for _ in $(seq 1 60); do kill -0 "$SERVER_PID" 2>/dev/null || return 0; sleep 1; done
    kill -KILL "$SERVER_PID" 2>/dev/null || true
}
trap cleanup INT TERM EXIT
```

Please verify it with `systemctl --user stop`, not a bare `kill`. systemd signals the whole control group and may mask the leak; that changes how bad it is, not whether it needs fixing.

**2. Both failure paths kill a healthy server.** Health timeout and warmup failure both do `cleanup; exit 1`. Measured boot-to-health on this box was 35, 30, 125 and 126 s against your 30 × 5 s = 150 s default, and those cold ones were with a warm page cache and the checkpoint already resident. A first boot after a host reboot reads a 27B checkpoint off disk on top of the 125 s. Log and fall through to `wait` on both paths, and raise the default window well past 150 s. A warmup pass that fails against a server answering `/health` is a reason to log, not a reason to take the server down.

**3. "Optional" is not what the diff ships.** `WARMUP=${WARMUP:-1}`, the README row says `WARMUP | 1`, the unit gains `Environment=WARMUP=1`, and `ExecStart` moves off `start_qwen.sh` onto the wrapper. That is every boot paying the 24–25 s warmup and every boot running through an ExecStart that can orphan the engine. Either default `WARMUP=0` and leave `ExecStart` where it is, or re-title this as the boot-path default flip it is and argue it on its merits.

**4. The correction I owe you: the JIT half of your motivation is not dead on the path production runs now.** I closed #48 at zero in-request compiles and told you the premise had evaporated. That zero was measured on `SPEC=dflash2 CTX=huge`. The int8 prefill work shipped since, and `docs/gotchas.md` gotcha 45's scope note now records that a current production boot (`CTX=fast`, `INT8_ACT=int8 PREFILL_ATTN=int8`) still logs three `JIT compilation during inference` lines — `_k_stats_kernel`, `_k_quant_kernel`, `_prefill_attn_kernel`. Those are exactly the counts I asked you for, and the repo produced them, not you, so take the point: uncovered serving-path kernels exist today.

What that does *not* establish is that a post-boot bench is the right fix for them. This repo's precedent for that class is a prewarm patch — `patches/dflash2-prewarm.patch`, `patches/spec-sampler-prewarm.patch` — closing the gap inside `kernel_warmup()`, where it costs milliseconds and cannot orphan anything. So re-aim the case for *this* PR at what a bench genuinely buys that a prewarm cannot: gotcha 39's "boots, /health 200, dies on the first prompt" OOM shape, where the point is to spend the transient allocation under controlled conditions before a user's prompt does. That argument I will take.

One thing to set expectations on: the live server here is a different unit out of a different tree, so merging this changes repo defaults, not the running box. Lower risk, also lower urgency.

**Deadline, since your own week has already elapsed once: Friday 2026-09-20.** Push the trap fix, the two non-fatal paths and either `WARMUP=0` or a re-titled PR, and I will re-run it and merge. If nothing lands by then I will close it and write the cold-cache boot budget — 35 / 30 / 125 / 126 s against a 150 s window — into `docs/gotchas.md`, because that is the durable finding either way and it came out of your PR. Say the word if you would rather I just do that now.

The `exec`-means-no-post-boot-hook observation was right, `bench/warmup.sh` is carefully written, and the gotcha-45 scope note says you were not wrong about JIT either. It is the wrapper's process handling that is holding this up, and it is three lines.

mhenrichsen:
Merging. I pushed the fixes onto your branch (`2381e82`) rather than sending them back to you — this PR allows maintainer edits, the changes are mechanical, and there is no reason for you to be holding a deadline for three lines of mine.

**What you got right, and what the GPU said.** I ran your branch against the 0.28.0 production stack (`SPEC=dflash2 CTX=fast PREFIX_CACHE=1 DFLASH_TOKENS=15 INT8_ACT=int8 PREFILL_ATTN=int8`, scratch port, production unit stopped). The wrapper works: boot-to-health 36-50 s, warmup pass 21-25 s, `[warmup] done` / `[server] Warmup OK`. And the motivation is live — the two bench passes absorbed exactly the three kernels gotcha 45's scope note names:

```
jit_monitor.py:141 Triton kernel JIT compilation during inference: _k_stats_kernel
jit_monitor.py:141 ... _k_quant_kernel
jit_monitor.py:141 ... _prefill_attn_kernel
```

Those compiled inside `vllm bench serve` instead of inside someone's first prompt. That is the thing worth having. The `BENCH=(...)` array, the OPENAI_API_KEY precedence chain and the `-fast` model re-derivation all match `start_qwen.sh` — that was careful work. I also checked the eight `vllm bench serve` flags you pass against 0.28.0 (`vllm bench serve --help=all`), since a flag vanished on us between releases once before (#66): 8/8 still present.

**Where I was wrong, and it's on the record.** My Sep 2 and Sep 13 comments said the non-blocking `wait` in `cleanup()` leaves a 23 GB orphan. It does not, on this box. Measured both ways on your branch unmodified:

- `systemctl --user stop` on a transient unit: returned in under a second; +25 s later `pgrep -af "[v]llm serve"` empty, card at 18 MiB.
- bare `kill -TERM` to the wrapper pid only, 20 ms polling: `wrapper_exit_s=0.651  server_exit_s=0.651`, log tail `Application shutdown complete`, no survivors, 18 MiB.

So the real cost is about 0.36 s of ordering slack against the unit's `ExecStartPre` VRAM gate, not a leak. I still took the blocking `cleanup()` + `trap ... EXIT` as cheap hardening — a wedged engine could outlive a 0.65 s shutdown — but it is hardening, not the defect I claimed.

**The defect that is real is the fall-through, and it is worse than a transient bench failure.** With `bench/warmup.sh` moved aside, your branch under a unit with `Restart=on-failure`:

```
[server] HTTP 200 OK
[server] warmup failed
ExecMainStatus=1 ; ActiveState=activating (auto-restart) ; NRestarts=2
```

A full 23 GB boot, killed one second after `/health` returned 200, on loop. The health-timeout branch does the same thing mid-boot. One correction to my own framing: the trigger I expected (`warmup.sh`'s hard-coded `models/…` and `venv/bin/vllm` not resolving in a deploy tree) does *not* fire here — both trees resolve. So it has to be justified on transient failures, a 401, or flag drift. Either way a server answering `/health` is more useful than a restart loop.

**What I pushed** (`2381e82`):

1. `cleanup()` forwards TERM, polls 60 s, then KILL; `trap cleanup INT TERM EXIT`.
2. Health timeout and warmup failure log and fall through to `wait "$SERVER_PID"`. `WARMUP_ATTEMPTS` 30 -> 120; a boot that never completes is `TimeoutStartSec`'s job.
3. `WARMUP` defaults to 0 and the unit keeps `ExecStart=start_qwen.sh`. The PR says "optional" four times; the unit diff made it the default boot path for every deployment. Opt-in is what the text claims and it is the cheaper thing to land. If we want it on by default that is a separate change, argued as a boot-path default flip.
4. Gotcha citations: 39 -> 35, 50 -> 45. Not your error — `231592a` ("58 entries down to 53") renumbered them under you after the branch point. Gotcha 8 was deleted rather than renumbered, so that reference is dropped.
5. Header and README re-aimed at gotcha 35's OOM shape plus gotcha 45's scope note, and the README now carries the 21-25 s cost next to the `WARMUP` row.

Re-verified after the fixes: stop and bare SIGTERM both leave 18 MiB, server exits strictly before the wrapper (0.662 s vs 1.024 s), warmup-missing gives `warmup failed, serving anyway` and the unit stays `active` with `NRestarts=0`, and `WARMUP` unset produces no warmup at all.

**Still unproven, and I am saying so rather than pretending otherwise:** every boot number above is a warm torch.compile cache. I did not clear the cache, so I did not test a cold boot against the health budget — at the 125-126 s cold figures we have on record, 120 x 5 s leaves plenty of room, but the old 30 x 5 s would have been at ~84% of budget. I also did not test a non-systemd or WSL2 host, where `KillMode` is irrelevant and a genuine orphan could still exist; no such box here.

Follow-up I'll file separately, not a blocker: `bench/warmup.sh` duplicates `start_qwen.sh`'s MODEL and API-key resolution and will drift.

Thanks — the wrapper shape and the warmup passes are yours and they're staying.

PR #56, opened 2026-09-02
fix(single-user): alternative.sh exits silently under set -e when REQ_METRICS is unset

linear3735:
`bash single-user/alternative.sh` (the documented 256k int4-KV path) exits
silently on a clean checkout — no output, exit 1 — so the server never starts.

Root cause: the script runs with `set -e`, and

    METRICS_ARGS=$([ "${REQ_METRICS:-0}" = 1 ] && echo --enable-per-request-metrics --enable-force-include-usage)

returns 1 when `REQ_METRICS` is unset (the `[ ... ]` test fails and the `&&`
short-circuits), so the command substitution fails and `set -e` terminates the
script right before `exec vllm serve`.

The neighbouring `ASYNC_ARGS` line already uses the safe `|| echo ...` fallback;
this change adds `|| true` to `METRICS_ARGS` for the same behaviour
(empty by default, set when `REQ_METRICS=1`).

Repro before fix:

    bash -x single-user/alternative.sh   # stops right after `+ METRICS_ARGS=`

### DIFF
--- a/single-user/alternative.sh
+++ b/single-user/alternative.sh
@@ -68,7 +68,7 @@ ASYNC_ARGS=$([ "$ASYNC_SCHED" = 1 ] && echo --async-scheduling || echo --no-asyn
 
 # REQ_METRICS=1: per-request timing fields + usage on every response (issue #51).
 # Not with --disable-log-stats (the timing fields need the engine-stats path).
-METRICS_ARGS=$([ "${REQ_METRICS:-0}" = 1 ] && echo --enable-per-request-metrics --enable-force-include-usage)
+METRICS_ARGS=$([ "${REQ_METRICS:-0}" = 1 ] && echo --enable-per-request-metrics --enable-force-include-usage || true)
 
 exec vllm serve "$MODEL" \
   --served-model-name qwen3.8-27b \


cpuchip:
Confirming this on a second platform: RTX 3090, native Ubuntu 26.04, Python 3.14, clean
checkout of main at 453104e. `bash single-user/alternative.sh` with default environment exits
before any output; `bash -x` ends at `+ METRICS_ARGS=`, and the server log is zero bytes.

Isolated: `bash -c 'set -e; A=$([ "${X:-0}" = 1 ] && echo y); echo reached'` returns 1 and
never prints. With `|| true` on the line, the same launcher reaches `exec vllm serve` with
default environment.

Two scoping notes from the same check: `ASYNC_ARGS` on the line above is not affected (its
`|| echo --no-async-scheduling` fallback always exits 0), and `start_qwen.sh` carries the same
`METRICS_ARGS` line but does not use `set -e`, so it is not affected either. The fix here is
the right one and the only line that needs it.

mhenrichsen:
You are right about both the failure and the mechanism, and thank you for the `bash -x` repro — "stops right after `+ METRICS_ARGS=`" is the whole diagnosis in one line.

This landed on main a few hours before your PR, as `6ceb64a` (#59, reported independently by @kucheryavenkovn), so I am closing this as already fixed rather than merging. It is my regression; sorry for the lost boot.

One note on the two fixes, since they are not quite equivalent. Yours adds `|| true`, matching the neighbouring `ASYNC_ARGS` line. Main instead uses an array:

```bash
METRICS_ARGS=()
[ "${REQ_METRICS:-0}" = 1 ] && METRICS_ARGS=(--enable-per-request-metrics --enable-force-include-usage)
...
  "${METRICS_ARGS[@]}" \
```

Both cure the `set -e` exit. The array additionally survives word-splitting: an unquoted `$METRICS_ARGS` expands correctly today only because neither flag contains a space, which is a property of the current flag list rather than of the code. The top-level `[ … ] && …` list is itself exempt from errexit, so no `|| true` is needed there. `start_qwen.sh` and `batch/start_qwen.sh` got the same treatment — they lack `set -e`, so they never crashed, but they had the same latent quoting issue.

The `ASYNC_ARGS` line you pointed at is still the `$( … || echo … )` form. It is safe (the fallback always succeeds) but has the same quoting shape, so I will convert it to an array in a follow-up.

Since you are clearly running `alternative.sh` on a clean native checkout: if you have any interest, #57 wants exactly that setup, and #60 is an unreproduced 128k decode regression where a second bare-metal data point would be worth a lot.

PR #57, opened 2026-09-02
Fix the int4 KV speculative-attention scratch-buffer sizing from #46: buffers were sized in sequences but indexed in tokens

cpuchip:
This fixes a defect in the attention extension we contributed in #46 for the int4 KV cache
(`spec-decode-int4-kv-mq3d.patch`, the `int4_per_token_head` path only; fp16, int8 and fp8
KV are not affected). It is not a bug in upstream vLLM. It is not a speedup at this repo's
default configuration (`max_num_seqs=1`, CUDA graphs on, draft depth 7, where draft depth is
the number of draft tokens proposed per step), measured on two cards, and the regimes where it could be (deeper
drafts, or more than one sequence) are untested for throughput here. What it does is make
two things agree that had drifted apart: the rule that decides which batches the split-KV
attention kernel accepts, and the size of the scratch buffers that kernel needs. Any future
disagreement is reported instead of silently routing the batch to the serial kernel. Numbers
and verification detail: [docs/spec-decode-scratch-token-units.md](docs/spec-decode-scratch-token-units.md).

What it carries:

1. **The fix** (`patches/spec-decode-scratch-token-units.patch`): the three scratch buffers
   behind the split-KV attention kernel are sized in query tokens, derived from the engine
   config, instead of in sequences.
2. **Reporting when it recurs**: an assert at construction; a counter of accepted batches that
   exceeded capacity ("breaches"); a flag set on the first breach and never cleared, readable
   through `mq3d_breach_state()`; and one diagnostic line rather than one per batch. Kernel
   selection is otherwise unchanged: the serial kernel remains the fallback.
3. **A property test** (`bench/mq3d_capacity_property.py`, run with `python`): checks the
   sizing rule over configurations the hardware cannot boot, with two negative controls that
   fail on the pre-fix sizing.
4. **The independent numerical check** (`bench/mq3d_layer2_oracle.py`) with its verdict
   records (`bench/mq3d_layer2_verdicts.jsonl`), written on a second machine without
   reference to this patch's code, so it can be rerun.
5. **A null result**: no throughput change at draft depth 7, with the arithmetic for why.

The branch also carries the one-line `alternative.sh` fix that #56 proposes (the same `|| true`; we
hit the same silent exit on a native checkout), so that it boots on its own. Once #56 merges, that
commit rebases to nothing.

## What was wrong

Definitions first. vLLM's Triton attention has two decode kernels: a serial one that walks
the whole KV cache in one pass (the code calls it 2D), and a split-KV one that divides that
walk into segments, runs them in parallel, and reduces the partial results (3D). The 3D kernel
keeps those partial results in three scratch buffers, indexed by query token. Speculative
decoding (here DFlash2, the draft model this repo ships) verifies several draft tokens per
sequence in one pass, so a verify batch of `depth + 1` tokens carries more query tokens than
it has sequences; the code calls that multi-query.

Upstream sizes the scratch buffers by `seq_threshold_3D`, a count of sequences equal to
`128 // num_heads_kv`, which is 32 for Qwen3.8-27B (4 KV heads). With CUDA graphs on, it
replaces that with the entry of the graph capture list closest to it (by absolute
difference), so that each captured graph covers one execution path; at `max_num_seqs=1` the
list is `[1, 2, 4, 8]`, so the buffers get 8 rows. In upstream this is correct, because
upstream never sends a multi-query batch to the 3D kernel (`or max_seqlen_q > 1` routes it to
2D), and on that path one sequence is one token.

Our #46 extension opened the 3D kernel to multi-query verify batches of up to 16 tokens per
sequence. It knew the buffers were indexed by token (its own comment says the one-token
equivalence "is what the old num_seqs check silently relied on for scratch-buffer capacity"),
but it added a runtime capacity check rather than resizing the allocation. The check kept the
path memory-safe, but it made the failure invisible. At this repo's default, the extension
accepted batches of up to 16 tokens while the buffers held 8 rows, so an accepted batch of 9
to 16 tokens quietly took the 2D kernel, and nothing logged it. In our runs only 9-token
batches occurred; 10 to 16 are covered by the property test and the independent check's
16-token case, not by live observation. At larger `max_num_seqs` the sequence-unit sizing is
short even without CUDA graphs, since `max_num_seqs x 16` tokens exceeds `seq_threshold_3D`
rows.

Found while evaluating whether the #46 kernel should be enabled by default.

## What changes

One variable was doing three jobs in three units. It becomes three:

- `seq_threshold_3D`: sequences, the 3D-eligibility threshold, unchanged, including the
  capture-list substitution.
- `MAX_QUERY_LEN_3D` (16): query tokens per sequence, the single source for both the
  eligibility check and the allocation.
- `scratch_token_capacity_3d`: total query tokens, the allocation:
  `min(max_num_batched_tokens, min(max_num_seqs, seq_threshold_3D) * MAX_QUERY_LEN_3D)`,
  where `seq_threshold_3D` is the value after the substitution, the same value the
  eligibility check uses. An accepted batch has at most `min(max_num_seqs, seq_threshold_3D)`
  sequences of at most 16 tokens, so it fits by construction whatever the substitution picks.

The capacity is derived from the engine config at construction and printed in the boot log
with its inputs and byte cost. Construction asserts that all three buffers carry it. If an
accepted batch ever exceeds capacity, the counter and flag above record it and one diagnostic
is logged; the 2D kernel serves the batch.

Cost for Qwen3.8-27B-W4A16-AutoRound with its drafter, at `max_num_seqs=1`: 16 token rows
across 4 allocations (one per attention backend instance, not per layer), 28.22 MiB, 14.11
MiB more than before. The formula grows with `max_num_seqs`; its ceiling at
`seq_threshold_3D=32` is about 903 MiB for this model, printed rather than capped, because a
silent cap would recreate the fallback this removes.

What it does not touch: the capture-list substitution itself. Changing which captured graph
a batch size replays is a separate change with its own tests.

## Measured

Single-stream decode tokens per second on an RTX 3090, generating 384 tokens after a NASA
history prompt, boots alternating patched and unpatched, dispatch instrumentation off during
timing (which kernel ran was proven in separate instrumented boots). At a 6,747-token prompt,
2 boots per arm: +0.18%, against within-arm spreads of 0.36% and 0.09%. With the
instrumentation on, 3 boots per arm: +0.13% against a 0.54% spread. At 68,013 tokens, 2 boots
per arm: within 0.2%, the reporting resolution at that sample size. On an RTX 4090 with a fresh compile cache per boot, where fresh compiles of the same code change draft
acceptance enough that tok/s is not an arm metric, the per-decode-step time differs by less than 0.5% in
every pair (26.0 and 38.1 ms per step at 6.7k and 68k). No throughput effect at draft depth 7 on either card. The band this
fixes occurred in 1 of 152 decode steps in the 68k run (a 9-token batch, hitting all 16 int4
attention layers of that step), because the ordinary verify batch at depth 7 is 8 tokens and
8 already ran the 3D kernel. The drafter is trained at depth 7, and the deeper regime is not
reachable with it on this card: at depth 11 the engine failed to boot in all eight attempts,
at both 262,144 and 98,304 context and with and without a prefix-match unit, on the KV-cache
coordinator's block-divisibility assert or, on at least one attempt, out of device memory
during graph capture; and the drafter itself clamps to 7 proposed tokens and fills the rest
of the verify batch from context. So the deeper regime is untested for throughput. Conditions, the dispatch table and the cost
table are in the linked doc.

## Verified

- Property test: 238,276 accepted batches over 4,032 configurations, 0 violations. The two
  pre-fix sizings fail it: sequence-unit sizing 129,498 times, and sequence-unit sizing after
  the capture-list substitution (what #46 ran under CUDA graphs) 140,008 times.
- Independent numerical check (included, see item 4), written on the 4090 box without
  reference to this patch, against a separate fp32 reference: 10 of 10 cases, and 10 of 10
  again when rerun on the 3090 from this branch with identical per-case values, including a 17-token single sequence
  (rejected) against a 16+1 two-sequence batch (the same 17 tokens, accepted). Reverting to
  sequence-unit sizing makes every multi-query case breach and leaves exactly the one-token
  and 17-token cases green.
- Live at `max_num_seqs=8`: 64 tokens across 8 sequences served by the 3D kernel (function
  check only; pre-fix not run at this setting, no throughput measured).
- Mutations: sequence-unit allocation refuses to boot at the assert; capacity shrunk to 8:
  breaches, with an exact count (1 after the first, 16 at the end) and one diagnostic.
- The full patch set applies clean to the stock vLLM 0.27.1 wheel in the Dockerfile's order
  (an apply check; the same set booted and served for every number above).

## Known nit

The boot log states the capacity formula as a sentence; the inputs and result are printed and
recomputable, but the sentence could go stale. Left as a follow-up rather than folded in
because the independent check verified these exact bytes.


mhenrichsen:
Merged as `c32be47`. This is the most carefully argued PR this repo has had: you found a defect in your own contribution, you say plainly that it is not a speedup, you distinguish what you observed live from what only the property test covers, and you had a second seat write the numerical check without reading your patch. The "known nit" at the end is the right instinct too.

Here is what I ran on the reference 3090 before merging, including one result that adds to yours and one correction to something I nearly reported.

## Confirmed

The patch applies clean to the deployed tree. Your property test reproduces exactly: 238,276 policy-eligible batches over 4,032 config tuples, 0 violations, with the limiter split 1,341 / 2,691 as you report.

**Dispatch census, `max_num_seqs=8`, `VLLM_INT4_MQ_3D=1`, one bench run each** (`alternative.sh`, int4 + TRITON_ATTN, DFlash2 depth 7, `VLLM_INT4_MQ_3D_DEBUG=1`):

| | 3D | 2D |
|---|---:|---:|
| stock | 440 | 1304 |
| patched | 752 | 896 |

312 batches move off the fallback. The boot line reads `capacity=128 query tokens (min of max_num_batched_tokens=2048, min(max_num_seqs=8, seq_threshold_3D=32) * max_query_len_3d=16)`, `48.38 MiB for this instance`, and at `max_num_seqs=64` it reads 512 tokens / 193.50 MiB — both matching your formula. No `DEGRADED` in any boot. The reason bits are genuinely useful: the 896 remaining 2D batches all report `max_query_len_policy>16`, i.e. rejected by policy before the capacity question, which is what makes the census readable at all.

## Your throughput null reproduces at concurrency too

You checked `max_num_seqs=8` for function only and said so. That was the gap I wanted to close, because with 8 sequences the band is reached deterministically (8 × 8 = 64 query tokens against sequence-sized rows) rather than in 1 step in 152. Interleaved boots, instrumentation off, 1024-in / 256-out random:

| | C8 | C1 |
|---|---:|---:|
| stock | 170.96, 176.56 | 100.25, 100.39 |
| patched | 175.11, 176.65 | 100.28, 100.45 |

A 1.2% gap between arm means, inside a 3.3% within-arm spread. Null, consistent with your depth-7 single-stream result. So the 3D kernel is not faster than 2D at these shapes even when the fix lets it run — which is worth knowing before anyone argues for default-on.

**A correction I owe the record:** in the *instrumented* boots the arms read 70.81 vs 77.29 tok/s, a tidy-looking +9.1%. That is an artifact. The debug line prints once per attention call, and the arms make different numbers of calls (1,744 vs 1,648), so the arm that falls back more also prints more. I nearly reported it. Instrumented boots cannot be a throughput arm here, which is a stronger version of the point your own methodology already makes by timing with instrumentation off.

## One finding to follow up: the allocation is outside the memory budget

At `max_num_seqs=64` the KV pool is byte-identical between arms — 346,404 tokens, 7.23 GiB available, across 8 boots, not a token of difference. That is not the reassuring result it looks like. It means the extra rows are **not** coming out of `gpu_memory_utilization`: the metadata builder is constructed after the memory profile that sizes the pool, so the buffers land in the non-KV headroom, unaccounted. Resident memory confirms it: 23,121 MiB stock against 23,303 MiB patched at `max_num_seqs=64`, +182 MiB that the budget never sees.

That is fine at the sizes we ship and both arms boot at 180k context, so it does not block the merge. But this repo's characteristic OOM is a headroom OOM, not a pool OOM (gotcha 39, and the "boots, /health 200, dies on the first prompt" shape), and your formula's ceiling is ~903 MiB. An unaccounted 903 MiB is exactly the wrong kind of invisible. Two options, either fine by me: derive and allocate before the profiling run so the pool absorbs the cost honestly, or subtract the declared capacity from the KV budget explicitly. Your instinct not to silently cap is right; the fix is to make the budget see it, not to hide it.

## Side finding, unrelated to your patch

Chasing a way to hit the 9..16 band at `max_num_seqs=1`, I tried `DFLASH_TOKENS=15` — the LABD setting this repo recommends for single-user — on `alternative.sh`. It does not boot on the int4 path at all: `AssertionError` on `scheduler_block_size % hash_block_size` at 32k context, and at the default 256k it cannot fit the pool (needs 5.67 GiB, has 4.55). `DFLASH_TOKENS=7` on the same config boots fine. `DFLASH_TOKENS=15` is only documented against the KVarN launcher, so this is a documentation gap rather than a regression, and it is the same assert your depth-11 attempts hit. I will file it separately.

## Housekeeping

I dropped your `alternative.sh` commit in the merge. #59 fixed that same silent `set -e` exit on main a few hours earlier with an array rather than `|| true` — equivalent for errexit, and additionally immune to word-splitting. #56 proposed the identical fix; you both hit it the same day, which says the launcher was broken for everyone on a native checkout.

mhenrichsen:
Follow-up for @cpuchip, since this is where the depth-11 attempts died: that was not your patch and not the scratch buffers — it was the same bug as #63, now fixed in `dfee877`.

The drafter's KV block was promoted until its *page covered* the primary page, but the scheduler additionally requires every group's block to *divide* its LCM. On `int4_per_token_head` those two coincided by luck at `DFLASH_TOKENS=7` (848 against a primary 1696, exactly half). Above 7 the mamba page grows with the spec-decode state, the primary block moves off 1696, and the covering block stops dividing it — hence the block-divisibility assert at both 262,144 and 98,304 context, which is exactly the `MAX_LEN`-independent signature you saw.

Depth > 7 now boots on the int4 path, so the untested 9..16 query-token regime this PR's 3D scratch sizing was written for is reachable: `DFLASH_TOKENS=15` on `alternative.sh` gives a 53,908-token pool at `MAX_LEN=32768` and generates correctly. Worth noting the pool cost before you plan a sweep — 53,908 against 142,843 at depth 7 — so keep `MAX_LEN` modest in those arms.

cpuchip:
Depth 15 on current main (dfee877), the band this PR's sizing was written for, measured on the RTX 4090 (WSL2, the container image built from main), `alternative.sh` int4, `DFLASH_TOKENS=15`, `MAX_LEN=32768`, `MAX_SEQS=4`, `VLLM_INT4_MQ_3D=1`, `VLLM_INT4_MQ_3D_DEBUG=1`:



- Boots with #63 in: pool 43,866 tokens at 32,768 max length on this card (your 53,908 on the 3090 has a different budget). The scratch declares capacity 64 query tokens (min(max_num_seqs=4, seq_threshold_3D=32) x max_query_len_3d=16).

- Dispatch census over two tool-calling conversations (flightbench mission, ~60 chat completions each) and one judgment run: 712 verify-shaped batches (max_seqlen_q <= 16), all 712 on the 3D path, 0 capacity fallbacks, breach_count 0, no DEGRADED line. Every width in the 9..16 band is present: 4 sequences at 10, 12, 14 and 16 query tokens (32 batches each at 10/12/14, 64 at 16), plus 1..3 sequences at 16, and the 1..8 widths. The 4,223 prefill chunks (max_seqlen_q 1840) went 2D by the query-length policy, as designed.

- Quality, thinking off: judgment 8/8; mission 14/18 twice, identical misses, no repetition, no truncation. Same conversation with the 3D path off at depth 15 (`VLLM_INT4_MQ_3D=0`): judgment 6/8, mission 15/18 twice, identical misses, same wall (151 to 152 s against 150 to 152 s); one decision apart from the 3D rows, both clean, and the same batch-shape census. Same with the follow-up patch (scratch allocated inside the memory budget) applied at boot: two sets allocated at model load (24.19 MiB at 396,288 B/row and 32.25 MiB at 528,384 B/row, 56.44 MiB together, capacity 64), all four builders adopted, zero "AFTER the memory profile" lines; `Model loading took` 15.92 to 15.98 GiB, `Available KV cache memory` 4.81 to 4.75 GiB, pool 43,866 to 43,338 tokens (528 tokens at this depth's per-token cost), so the bytes now sit inside the budget instead of beside it. Quality at depth 15 with it: judgment 6/8, mission 16/18, no repetition, the same census (every eligible batch on 3D, no fallback, no DEGRADED line).



So at depth 15 the sizing is exercised end to end with no fallback, and the numbers this PR promised for the 9..16 band are now observed rather than derived. The follow-up patch (scratch inside the budget) is #67, with this depth-15 row in its body.

Issue #59, opened 2026-09-02
single-user/alternative.sh dies silently under set -e when REQ_METRICS is unset (restart loop, @201d237)


kucheryavenkovn:
## `single-user/alternative.sh` exits silently under `set -e` when `REQ_METRICS` is unset (restart loop)

Commit `201d237` added to all three launchers:

```bash
METRICS_ARGS=$([ "${REQ_METRICS:-0}" = 1 ] && echo --enable-per-request-metrics --enable-force-include-usage)
```

`single-user/alternative.sh` is the only launcher with `set -e` (line 11). With `REQ_METRICS` unset/`0` the `[ ... ]` test fails, the `&&` list short-circuits, the command substitution returns exit status 1, and `set -e` kills the script **silently** — right before `exec vllm serve`, with zero output. A container whose entrypoint runs `alternative.sh` (e.g. the int4 KV-cache profile) goes into a Docker restart loop: `verify.sh` passes, `[int4-start] exec single-user/alternative.sh`, immediate exit, repeat.

`batch/start_qwen.sh` and `single-user/start_qwen.sh` have the same line but no `set -e`, so they survive (empty `METRICS_ARGS`).

Reproduced on `sha-453104e`, WSL2, RTX 4090: 12 restarts, exit before any vLLM output; with `REQ_METRICS=1` the script proceeds normally.

Suggested fix:

```bash
METRICS_ARGS=$([ "${REQ_METRICS:-0}" = 1 ] && echo --enable-per-request-metrics --enable-force-include-usage || true)
```

(Local workaround meanwhile: `sed -i 's/--enable-force-include-usage)$/--enable-force-include-usage || true)/' single-user/alternative.sh` before exec.)

mhenrichsen:
Correct in every detail, including the mechanism — thank you for the precise write-up. Fixed on main in `6ceb64a`, and closing this with it.

The failure is exactly as you describe: a simple assignment whose command substitution exits non-zero is one of the cases `set -e` does kill, so `alternative.sh` died between `verify.sh` and `exec vllm serve` with nothing on stdout. I shipped the array form rather than `|| true`:

```bash
METRICS_ARGS=()
[ "${REQ_METRICS:-0}" = 1 ] && METRICS_ARGS=(--enable-per-request-metrics --enable-force-include-usage)
```

[...]

PR #65, opened 2026-09-02
launchers: ASYNC_ARGS and TOOL_ARGS as arrays (finishes the #59 conversion)

linear3735:
Continues the array conversion #59 started for `METRICS_ARGS`. Two command-substitution arg lists remained in the launchers:

- **ASYNC_ARGS** (`single-user/alternative.sh`, `single-user/start_qwen.sh`) was `$( [ ... ] && echo ... || echo ... )`. The `|| echo` fallback always exits 0, so it is errexit-safe and never killed a script — but the unquoted `${ASYNC_ARGS}` expansion is word-split, the same latent shape #59 removed from `METRICS_ARGS`.

- **TOOL_ARGS** (`single-user/start_qwen.sh`, `batch/start_qwen.sh`) was `$( [ ... ] && echo ... )` with no fallback — the exact shape #59 fixed. The substitution exits 1 whenever `TOOLS=0`, so it would kill a launcher under `set -e` today; it only survives because neither of those two scripts uses `set -e`. It also word-split `$TOOL_PARSER`.

Both are now arrays guarded by a top-level AND-OR list (errexit-exempt, the pattern #59 introduced) and expanded quoted with `"${...[@]}"`, matching `SPEC_ARGS` and `METRICS_ARGS`. `$TOOL_PARSER` is now a single quoted array element, so a parser name containing whitespace no longer splits.

## Verified

- `bash -n` on all three launchers.
- Under `set -e`: `ASYNC_SCHED` ∈ {unset, 0, 1} and `TOOLS` ∈ {unset, 0, 1} each expand to the expected argument list (no silent exit).
- An empty `TOOL_ARGS` expands to zero arguments (not one empty string), so the `exec vllm serve` line is unchanged when tools are off.

### DIFF
--- a/batch/start_qwen.sh
+++ b/batch/start_qwen.sh
@@ -105,7 +105,10 @@ fi
 # 0.27.1, which is the tool-side adapter of the same parser engine that
 # --reasoning-parser qwen3 already uses (vllm/parser/qwen3.py).
 TOOL_PARSER=${TOOL_PARSER:-qwen3_coder}
-TOOL_ARGS=$([ "${TOOLS:-1}" = 1 ] && echo --enable-auto-tool-choice --tool-call-parser $TOOL_PARSER)
+# Array, not $( [ ] && echo ): exits 1 when TOOLS is off (the shape #59 fixed)
+# and word-splits $TOOL_PARSER; the array keeps the parser as one element.
+TOOL_ARGS=()
+[ "${TOOLS:-1}" = 1 ] && TOOL_ARGS=(--enable-auto-tool-choice --tool-call-parser "$TOOL_PARSER")
 
 # REQ_METRICS=1: per-request timing fields + usage on every response (issue #51).
 # Not with --disable-log-stats (the timing fields need the engine-stats path).
@@ -192,5 +195,5 @@ exec venv/bin/vllm serve "$MODEL" \
   --reasoning-parser qwen3 \
   --enable-prompt-tokens-details \
   "${METRICS_ARGS[@]}" \
-  ${TOOL_ARGS} \
+  "${TOOL_ARGS[@]}" \
   ${EXTRA_ARGS}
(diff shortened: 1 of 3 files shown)


mhenrichsen:
Merged as 2bbd292. Verified on the reference 3090 before merging, in three stages.

**The argv is unchanged except where you intend it.** I stubbed `vllm` to print one argument per line and ran all three real launchers over the full matrix — `ASYNC_SCHED` and `TOOLS` each unset/0/1, `TOOL_PARSER` default and `"two words"` — 108 invocations per branch. The diff against main is 12 hunks, and every one of them is the same change:

```
-| two
-| words
+| two words
```

Nothing else moves. `TOOLS=0` still contributes zero arguments rather than one empty string, which was the thing worth checking about `"${ARR[@]}"` on an empty array.

**The errexit claim holds.** Injecting `set -e` at the top of the real scripts is not a valid test — both branches then die at rc=127 on unrelated preflight — so I tested the two shapes in isolation:

| shape | `TOOLS=0` | `TOOLS=1` |
|---|---|---|
| `TOOL_ARGS=$( [ ] && echo ... )` | **rc=1, script never reaches the end** | argc=3 |
| array + AND-OR guard | argc=0, reaches the end | argc=3 |

So your reading is exactly right: it is dormant only because neither of those two scripts sets errexit today, and the guard form is errexit-exempt because the failing command is part of an AND-OR list.

**It serves.** Production restarted on the new `single-user/start_qwen.sh`: healthy in 50 s, live cmdline shows `--no-async-scheduling` and `--enable-auto-tool-choice --tool-call-parser qwen3_coder` as separate argv entries, and a `tool_choice=auto` request round-trips `get_weather({"city": "Paris"})`.

That is the last command-substitution arg list in the launchers; `SPEC_ARGS`, `METRICS_ARGS`, `ASYNC_ARGS` and `TOOL_ARGS` are now all arrays. `VISION_ARGS`, `ATTN_ARGS`, `PREFIX_ARGS` and `EXTRA_ARGS` are still unquoted strings, but those are plain assignments rather than substitutions, so they carry the word-splitting half of the problem without the errexit half — `EXTRA_ARGS` deliberately, since users pass multiple flags through it. Not asking you to chase them; noting it so the next reader knows the sweep was scoped, not complete.

Thanks for finishing this off.

Issue #63, opened 2026-09-02
DFLASH_TOKENS=15 does not boot on the int4 path (alternative.sh): block-divisibility assert at 32k, KV shortfall at 256k


mhenrichsen:
`DFLASH_TOKENS=15` is the LABD setting this repo recommends for single-user serving (+50% on verbatim reproduction, +9% at C1; `docs/long-context.md`, gotcha 45). It does not boot on `single-user/alternative.sh`, the int4-KV path. Found while looking for a way to reach the 9..16 query-token band on the int4 3D attention path for #57; not a regression, and not caused by that PR.

Reference RTX 3090, `alternative.sh` (`--attention-backend TRITON_ATTN --kv-cache-dtype int4_per_token_head`), `SPEC=dflash2 MAX_SEQS=1`, launcher defaults otherwise:

| config | result |
|---|---|
| `DFLASH_TOKENS=15`, default `MAX_LEN=256000` | `ValueError`: needs 5.67 GiB of KV, 4.55 GiB available; "estimated maximum model length is 185840" |
| `DFLASH_TOKENS=15`, `MAX_LEN=32768` | `AssertionError` in `kv_cache_coordinator.py:84` — `assert scheduler_block_size % hash_block_size == 0` |
| `DFLASH_TOKENS=15`, `MAX_LEN=32768`, `PREFIX_CACHE=0` | same `AssertionError` |
| `DFLASH_TOKENS=7`, `MAX_LEN=32768` | boots, healthy in 120 s |

Two separate things:

1. **The 256k row is just memory** and behaves as documented — the verify block grows the per-request bill (gotcha 45's `102 MiB * (DFLASH_TOKENS + 2)`), so 256k context and a 16-token verify block do not both fit. Arguably the launcher should say so rather than let the engine raise, since `MAX_LEN` and `DFLASH_TOKENS` both default to values that are individually fine and jointly impossible on this path.

2. **The 32k row is the real one.** The block-divisibility assert fires with prefix caching on *or off*, and `DFLASH_TOKENS=7` on the identical config is fine, so it is the verify block width interacting with the scheduler/hash block sizes on the int4 path specifically. The same assert is what @cpuchip's depth-11 attempts hit in #57 ("the KV-cache coordinator's block-divisibility assert"), at both 262,144 and 98,304 context — so this is not about a particular `MAX_LEN`, and depth 11 and depth 15 are likely the same bug.

[...]

PR #67, opened 2026-09-03
int4 3D scratch: allocate inside the memory budget (#57 follow-up)

cpuchip:
The follow-up from the #57 merge review: the int4 3D scratch now allocates inside the memory budget.

1. `patches/spec-decode-scratch-within-budget.patch`: the attention `Impl` (constructed in `load_model`, inside the profiled window) allocates the scratch for its geometry into a pool; every metadata builder adopts it. `Model loading took` rises by the scratch bytes and `Available KV cache memory` drops by the same. The derivation moved into `mq3d_scratch_plan` and did not change; the property test is unaffected.
2. A builder that finds nothing to adopt still allocates, and logs it at WARNING as bytes outside the KV budget, so the unaccounted case cannot recur silently.
3. `bench/mq3d_scratch_pool_test.py`: the pool logic on CPU tensors (share, adopt-smaller, replace-larger, exclusive), with a negative control. Runs inside the image without a GPU.
4. Measurements and the reading in `docs/spec-decode-scratch-token-units.md`, "Follow-up".

Measured (details and conditions in the doc). Reference 3090, native, fresh compile cache per boot, `alternative.sh` int4, `MAX_LEN=180000`, patched minus stock at equal `MAX_SEQS`:

| `MAX_SEQS` | allocated at load | `Model loading took` | `Available KV` | KV tokens | free VRAM after boot | prompt |
|---|---|---|---|---|---|---|
| 8 | 112.88 MiB | +113 MiB | -164 MiB | -7,798 | +280 MiB | ok |
| 16 | 225.75 MiB | +225 MiB | -266 MiB | -12,347 | +462 MiB | ok |
| 32 | 451.50 MiB | +461 MiB | -471 MiB | -22,094 | +1,004 MiB | ok |
| 64 | 451.50 MiB | +461 MiB | -420 MiB | -20,144 | +824 MiB | ok |

At `MAX_SEQS=32` free VRAM after boot goes from 319 MiB to 1,323 MiB, out of the burst zone gotcha 39 describes, for about 22k of 230k KV tokens. That is the cost the review asked to have made visible.

A 4090 under WSL2 (`CTX=long`, k=7, compile cache disabled on every boot so each profiles cold) agrees: two stock boots identical to the token (184,701); patched at `MAX_SEQS` 4 / 16 / 32 drops the KV pool by 2,035 / 8,141 / 16,791 tokens (55 / 219 / 452 MiB) against 56.44 / 225.75 / 451.50 MiB allocated at load, model load +0.06 / +0.21 / +0.44 GiB, every boot 4 adopted and 0 after, every arm alive through a 6.7k and a 68k prompt. Resident memory is not an oracle for this: the caching allocator serves the scratch from segments the profiler already reserved, so `nvidia-smi` under-reads it; the KV line is the measure.

Two things to know:

- Builders of one geometry share a set (four builders, two geometries on this model), which is safe because an attention call writes and reduces the scratch within the call on one stream, the assumption the sharing across a group's layers already makes. Under ubatching (DBO) a group's builders run on separate streams, so a builder then adopts only a set no builder has adopted, and the rest allocate their own, logged. `use_ubatching` is off in every launcher here.
- Geometry is taken from the model config, as the builder derives it, not from the layer's own arguments; they differ for the drafter, and the first cut of this patch keyed on the layer's numbers and left the drafter's builder outside the budget. The boot log shows which path every builder took.

Depth 15 (the 9..16 band this sizing exists for, bootable since #63): 712 of 712 verify-shaped batches on the 3D path with no fallback at capacity 64; with this patch the two sets (56.44 MiB) allocate at model load, all four builders adopt, and the pool moves 43,866 to 43,338 tokens on the 4090 at `MAX_LEN=32768`. Rows and census in the doc's "Depth 15" section.

### DIFF
--- a/bench/mq3d_scratch_pool_test.py
+++ b/bench/mq3d_scratch_pool_test.py
@@ -1,0 +1,45 @@
+"""The int4 3D scratch pool (spec-decode-scratch-within-budget.patch), on CPU tensors: a model-load
+allocation is adopted by builders, builders of one geometry share it, a smaller derived capacity
+adopts the larger pool, a larger one allocates and replaces, an exclusive (ubatching) builder never
+shares, and a different geometry is a different key. Ends with a negative control: with the
+exclusive flag off, the second builder DOES share, which is the case the guard exists to prevent.
+
+Runs inside the image, no GPU:  /app/venv/bin/python bench/mq3d_scratch_pool_test.py
+Exit 0 = every assertion held (assert raises otherwise)."""
+import logging, sys
+import torch
+import vllm
+import vllm.v1.attention.backends.triton_attn as m
+logging.getLogger("vllm").setLevel(logging.ERROR)
+P = m.Mq3dScratchPlan
+def plan(cap, heads=4, hd=64):
+    row = 4 * (heads * m.NUM_PAR_SOFTMAX_SEGMENTS * hd + 2 * heads * m.NUM_PAR_SOFTMAX_SEGMENTS)
+    return P(num_heads_q=heads, num_heads_kv=2, headdim_padded=hd, seq_threshold_3D=32,
+             max_query_len_3d=16, capacity=cap, row_bytes=row, max_num_batched_tokens=2048, max_num_seqs=4)
+dev = torch.device("cpu")
+pool = m._MQ3D_SCRATCH_POOL; pool.clear()
+a = m.mq3d_scratch_acquire(plan(64), dev, "model load")
+assert a.origin == "model load" and a.adopters == 0 and len(pool) == 1, "load allocates, pooled"
+b = m.mq3d_scratch_acquire(plan(64), dev, "metadata builder")
+c = m.mq3d_scratch_acquire(plan(64), dev, "metadata builder")
+assert b is a and c is a and a.adopters == 2, "two builders share the load-time set"
+d = m.mq3d_scratch_acquire(plan(32), dev, "metadata builder")
+assert d is a and d.capacity == 64, "smaller derived adopts the larger pool"
+e = m.mq3d_scratch_acquire(plan(128), dev, "metadata builder")
+assert e is not a and e.capacity == 128 and pool[next(iter(pool))] is e and e.adopters == 1, "larger derived allocates and replaces"
+pool.clear()
+x = m.mq3d_scratch_acquire(plan(64), dev, "model load")
+y = m.mq3d_scratch_acquire(plan(64), dev, "metadata builder", exclusive=True)
+z = m.mq3d_scratch_acquire(plan(64), dev, "metadata builder", exclusive=True)
+assert y is x and z is not x and z.capacity == 64 and pool[next(iter(pool))] is x, "exclusive: first adopts, second allocates its own, pool keeps the first"
+w = m.mq3d_scratch_acquire(plan(64), dev, "metadata builder", exclusive=False)
+assert w is x, "non-exclusive still shares"
+g = m.mq3d_scratch_acquire(plan(64, heads=8), dev, "metadata builder")
+assert g is not x and len(pool) == 2, "different geometry is a different key"
+print("POOL UNIT TEST OK: 9 assertions")
+# negative control: the exclusive guard removed -> sharing across exclusive builders would be wrong
+pool.clear()
+x = m.mq3d_scratch_acquire(plan(64), dev, "model load")
+m.mq3d_scratch_acquire(plan(64), dev, "metadata builder", exclusive=True)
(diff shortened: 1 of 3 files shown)


mhenrichsen:
Merged as `e0fc19d`. This is the follow-up the #57 review asked for and it does what it says, verified in five arms on the reference 3090 before merging. It also turned up a measurement hazard that I think matters more than the patch, and one repo-level defect I fixed on top rather than sending back.

## The mechanism, in every arm

`alternative.sh`, int4, native venv, `MAX_LEN=180000` unless noted:

| arm | allocated at load | `Model loading took` | builders adopting | scratch WARNINGs |
|---|---|---|---|---|
| `MAX_SEQS=8` | 48.38 + 64.50 = **112.88 MiB** | 15.61 → **15.72 GiB** (+113 MiB) | 4 of 4 | 0 |
| `MAX_SEQS=32` | 193.50 + 258.00 = **451.50 MiB** | 15.61 → **16.06 GiB** (+461 MiB) | 4 of 4 | 0 |
| `DFLASH_TOKENS=15`, `MAX_SEQS=1`, 32k | 6.05 + 8.06 = 14.11 MiB | 15.61 → 15.63 GiB | 4 of 4 | 0 |
| `CTX=long` (int8, `start_qwen.sh`) | 24.19 + 32.25 = 56.44 MiB | — | 4 of 4 | 0 |

The model-load figure rises by exactly the bytes allocated, in both directions of scale, which is the claim. Two geometries, one set each, four builders adopting — no builder ever took the WARNING path. Every arm served correctly.

`CTX=long` is worth a note for anyone reading the table later: it pins `--kv-cache-memory`, so its pool stays at exactly the documented **138,696 tokens** either way. There the patch moves the accounting without moving the pool.

## The KV-pool numbers reproduce — but only paired correctly, and this is the hazard

`Available KV cache memory` is **not** reproducible boot to boot on this box at a fixed config. Two identical stock `MAX_SEQS=8` boots:

| boot | consumed (weights + non-torch) | peak activation | Available KV | KV tokens |
|---|---|---|---|---|
| stock `MAX_SEQS=8`, first | 16.03 GiB | **1.34 GiB** | 5.02 GiB | 243,682 |
| stock `MAX_SEQS=8`, repeat | 16.03 GiB | **0.42 GiB** | 5.94 GiB | **288,519** |

A 0.92 GiB swing, ~45,000 KV tokens, same config, same 18 MiB idle GPU — **larger than every delta in your table**. The whole difference is in peak activation, and it is the compile cache: I re-ran the patched `MAX_SEQS=8` arm changing one thing, a throwaway `TRITON_CACHE_DIR` + `VLLM_CACHE_ROOT`, and peak activation went 0.42 → 1.34 GiB, Available KV 5.78 → 4.86, pool 281,371 → 236,534, while `Model loading took` and consumed memory stayed byte-identical at 15.72 / 16.18 GiB. So a cold compile inflates the profiled peak and the pool absorbs it.

Your own harness disabled the compile cache on every boot, which is presumably why your stock boots were "identical to the token" — you controlled for it without needing to name it. Anyone repeating this without that control will measure the cache, not the patch. **Pair arms by peak activation, or read `Model loading took` / consumed memory, which are stable across all seven of my boots.**

Paired properly (both at 0.42 GiB peak), `MAX_SEQS=8`:

| | stock | patched | delta |
|---|---|---|---|
| Available KV | 5.94 GiB | 5.78 GiB | **−164 MiB** |
| KV tokens | 288,519 | 281,371 | −7,148 |
| free VRAM after boot | 1,103 MiB | 1,325 MiB | **+222 MiB** |

−164 MiB is your number exactly. My KV-token and free-VRAM deltas are a little smaller than yours (−7,148 vs −7,798, +222 vs +280), which is the kind of gap two different trees produce; the accounting identity holds in both.

## One thing I fixed rather than sending back

Merging this made `verify.sh` report a **false failure** on `spec-decode-scratch-token-units.patch`. Both patches still apply, in order, and both are required — but this one rewrites the region that one added, so its lines are no longer in the tree and neither the reverse dry-run nor the content check can find them. That is a stacking problem the repo will hit again, so I fixed it generally in `5259fc7`: a patch can declare `Supersedes: <basename>` in its header, honoured only when the declaring patch is itself applied. I added that line to your patch's header. Controls: with the declaring file removed, or with the declaration naming a different patch, the old FAIL comes back. `verify.sh` is now OK with 0 failures.

Also confirmed: `bench/mq3d_scratch_pool_test.py` passes 9 assertions with the negative control, and fails on an unpatched tree with `AttributeError: Mq3dScratchPlan` — a test that cannot fail is not a test, and this one can.

Thanks for the doc section. The "resident memory is not an oracle" point is the one I would have got wrong.

PR #69, opened 2026-09-03
kvarn: restore the sliding-window page padding hunk dropped in the 0.28.0 runner patch (stacked on #43)

cpuchip:
Stacked on #43: this branch is #43's current head (e34afaf) plus one commit.

1. **What it carries.** The second `attention.py` hunk of the 0.27.1 `kvarn/kvarn-v2-runner.patch`, restored inside `kvarn/kvarn-v2-runner-0.28.0.patch`. For KVarN cache dtypes on a hybrid model without skip layers it pads the DFlash drafter's sliding-window pages to the mamba page (`shared_page = cache_config.mamba_page_size_padded`). The attribute exists in 0.28.0 core. The condition is gated on the kvarn dtype, so it is a no-op for int4, int8 and bf16.

2. **What it fixes.** On #43 as filed, KVarN + DFlash2 collapses into repetition ("duct Register Register ...") from the second turn of a conversation, on two machines (an RTX 4090 under WSL2 and a native RTX 3090), at every draft width tried (1, 3, 7, 15), with prefix caching on or off. KVarN + MTP and KVarN with speculation off are unaffected, and the 0.27.1 fork passes the same cell 18/18 on both machines. The installed `kvarn_attn.py` is byte-identical between the two builds, so the regression is in what the runner feeds KVarN. A hunk-by-hunk census of the two runner patches found exactly one hunk missing, this one; it has been absent since the first 0.28 commit regenerated the patch.

3. **Why it breaks.** Without the hunk the drafter's sliding-window spec starts at a 16-token block with a 65,536-byte page. The block-promotion step then grows it to a 544-token block with a 2,228,224-byte page, which becomes the new maximum page for every group, and unification pads the KVarN target spec up to it (400,384 bytes more per manager block). Core allocation walks by that padded page while KVarN's packed view and flush lifecycle walk by seventeen 128-token tiles, so block ownership and byte ownership diverge and a later flush overwrites bytes the target still addresses as another block. With the hunk the drafter's page equals the mamba page from the start, it is excluded from promotion, and the two views stay identical. The boot-time line "Disabling fine-grained prefix-cache hits ... SlidingWindowManager" is a symptom of the same geometry and disappears with the fix.

4. **How it was verified.** Twelve-turn tool-using conversation (flightbench mission v6.1, thinking off, 900-token reply cap), KVarN k4v2 + DFlash2 k=7:
   - RTX 4090, WSL2: as filed 6-9/18 over three runs with 9-11 of 12 replies cut at the cap; hunk applied at boot 16/18 and 11/18 with no reply cut; this branch built on e34afaf 14/18 with no reply cut, and 14/18 with `PREFIX_CACHE=0`; int4 + DFlash2 on this branch 15/18, unchanged.
   - RTX 3090, native: as filed 3/18 with 10 of 12 replies cut; hunk restored 15/18 with no reply cut and none capped, and 15/18 with `PREFIX_CACHE=0`; KVarN + MTP 15/18 on the same box.
   - `bench/residue_sweep.py`, all 128 residues, KVarN + DFlash2 k=7 (RTX 4090): as filed, 46 of 46 residues broken before the run was stopped (coverage 0.00, mostly 2,695 characters of repetition, 2.7 minutes per residue); with the hunk, 0 of 128 broken, coverage 1.00, one identical 834-character answer on all 128.
   The full cell-by-cell record, including every knob that did not heal it: https://github.com/cpuchip/qwen38-27b-rtx3090/blob/v0.28-validation/docs/v0.28-validation.md

5. **Not in scope.** On the port, about half the turns of a tool-using conversation hit the tool-call cap on every tier, drafter or not (bf16 with speculation off: 9 of 12 turns), and prefix caching off removes it; not present on 0.27.1. Documented in the write-up as a separate regression.

### DIFF
--- a/.github/workflows/patch-integrity.yml
+++ b/.github/workflows/patch-integrity.yml
@@ -1,0 +1,28 @@
+name: patch integrity
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  git-apply:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check repository
+        uses: actions/checkout@v4
+
+      - name: Check pinned vLLM source
+        uses: actions/checkout@v4
+        with:
+          repository: vllm-project/vllm
+          ref: v0.28.0
+          path: .ci/vllm
+          fetch-depth: 1
+
+      - name: Validate patch series with git apply
+        run: bash patches/check_vllm_series.sh .ci/vllm
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ venv
 models
 api_key.txt
 bench/results/
+kv-persist-cache/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # Same stack as the README's venv install, frozen: Python 3.12 venv at /app/venv,
-# vLLM 0.27.1 (torch 2.13 / cu130 / Triton 3.7.1), every patch in patches/ applied,
+# vLLM 0.28.0 (torch 2.13 / cu130 / Triton 3.7.1), every compatible patch in
+# patches/ applied,
 # the KVarN KV cache installed, verify.sh --install run at build time.
 #
 # The base image is CUDA "base" + nvcc, not "devel": vLLM's wheels bring their own
@@ -25,7 +26,12 @@ RUN venv/bin/pip install -r docker/requirements.txt
 
 COPY . .
 RUN set -e; SP=$(venv/bin/python -c 'import vllm, os; print(os.path.dirname(vllm.__file__))' | tail -n1); \
-    for p in patches/*.patch; do echo "== $p"; patch -p1 -d "$SP" < "$p"; done; \
+    for p in patches/*.patch; do \
+      case "$p" in \
+        patches/dflash2-backport.patch) echo "== skip $p (DFlash2 is native in vLLM 0.28.0)"; continue ;; \
+      esac; \
+      echo "== $p"; patch -p1 -d "$SP" < "$p"; \
+    done; \
     bash kvarn/install.sh; \
     bash verify.sh --install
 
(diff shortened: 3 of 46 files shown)


cpuchip:
We should close this in favor of #43 the hunk lives there now too, thanks @alvins82

mhenrichsen:
Closing this because the hunk is in main, as part of #43 (`cd2fa94`) — but the finding is yours, and I want it recorded as yours rather than absorbed silently.

I spent two rounds on that failure and got as far as "DFlash2-on-KVarN is broken and `CTX=fast` on the same build is fine", which located the fault and explained nothing. You found the actual cause by doing the boring, correct thing: a hunk-by-hunk census of the 0.27.1 runner patch against the 0.28.0 one, exactly one hunk missing, absent since the first commit that regenerated the patch. The mechanism you wrote up — drafter page not pinned to the mamba page, so the promotion step grows it to a 544-token block whose page becomes the new maximum, core allocation then walks by the padded page while KVarN's packed view walks by seventeen 128-token tiles, and a later flush overwrites bytes the target still addresses — is the first explanation of that corruption anyone has produced, and it predicts the symptom I saw (character-exact on request 1, collapse afterwards) rather than merely accompanying it.

The evidence standard is worth calling out too: `residue_sweep` at 46 of 46 residues broken before the run was stopped, against 0 of 128 broken with the hunk, on two machines, with the "every knob that did not heal it" list written down. That is a stronger case than the one I built.

alvins82 landed the identical hunk on #43 as `6869c80` after your PR, so the two are byte-identical in the runner patch; I verified that before merging. What is left in this branch is the older base — the unrelated runtime and template changes #43 has since dropped — so there is nothing here to merge on top.

Confirmed on the reference 3090 with the merged tree: `CTX=huge` + DFlash2 reproduces an 800-character span of a 29.5k-token document 800/800 at both `DFLASH_TOKENS=7` and 15, where every previous revision scored 0/800.

Two things from your write-up I did **not** merge and would still like:

1. **The tool-call cap regression** — half the turns of a tool-using conversation hitting the cap on 0.28 on every tier, drafter or not, `PREFIX_CACHE=0` removing it, absent on 0.27.1. That is now a live bug in main, and it is not covered by anything I ran. Please open it as its own issue with the bf16 `SPEC=off` case as the minimal repro, since that one takes the drafter out of the picture entirely.
2. **The autotune finding on #43** — `merge_16x16_to_64x64_inverse_kernel` at 2 vs 4 warps deciding the whole trajectory, with the pin-one-at-a-time table proving it. That deserves to be an issue rather than a comment on a merged PR; it is the cleanest evidence I have seen that benchmark rows on this stack need a frozen autotune draw, and it belongs somewhere findable.

PR #77, opened 2026-09-04
docs: improve Windows/WSL2 onboarding

cofade:
## Summary

- add a sanitized `.env.example` with the recommended single-user settings and WSL2 pin-memory requirement
- show Linux/WSL and PowerShell first-run setup in the quick-start docs
- surface the `VLLM_WSL2_ENABLE_PIN_MEMORY=1` requirement before Docker startup
- document Windows WSL/Docker host-memory and `/mnt/c` venv workarounds

## Motivation

New Windows users can otherwise hit `RuntimeError: UVA is not available` during DFlash2 startup or host-memory/DrvFs failures during preparation. The template contains no secrets; `.env` remains gitignored.

The WSL2 guidance was tested against the Docker Desktop setup used by this repository.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,67 @@
+# Copy this file to .env before starting Docker Compose.
+# .env is ignored by git so local keys and tuning changes stay out of commits.
+
+# Speed-first single-user config for a local RTX 3090.
+# This is the repo's recommended default for best speed when the card is yours alone.
+SPEC=dflash2
+PREFIX_CACHE=1
+
+# Optional toggles kept commented out for easy activation later.
+# Uncomment ONLY when you need the behaviour below.
+
+# DFLASH_TOKENS=15
+# Use when the model is reproducing or quoting a long prompt/document/code block.
+# This is the "copy/edit/quote" mode: much faster when the answer is reusing context,
+# but it reduces request slots and context headroom compared with the default 7-token verify block.
+# Typical use: long docs, code edits, RAG with quoting, prompt-copy workloads.
+
+# CTX=long
+# Use when you need about 150k context and are willing to trade speed for more tokens.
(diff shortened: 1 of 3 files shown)


mhenrichsen:
Merged, with two corrections in `ccd714b` — both small, and the PR is a clear improvement on what was there.

**Verified rather than taken on trust:** `VLLM_WSL2_ENABLE_PIN_MEMORY` is a real 0.28.0 variable (`envs.py:307`, default `0`), and `platforms/cuda.py:310` reads it **only inside the WSL-detected branch** — on native Linux the function returns `True` before ever looking at it. So "native Linux can leave it enabled" is exactly right, and it is a genuine no-op there. Worth stating plainly because this README has burned itself on a plausible-looking pin-memory variable name before (`VLLM_WSL_PIN_MEMORY`, which does not exist); yours is the real one.

**What I changed:**

1. The template said `SPEC=dflash2 is the default`. It is not — `single-user/start_qwen.sh:146` is `SPEC=${SPEC:-mtp}`. The template *setting* dflash2 is the useful thing and I kept it; I only fixed the sentence, because someone debugging a profile would otherwise go looking in the wrong place.

2. `docs/docker.md` replaced `echo "VLLM_API_KEY=$(openssl rand -hex 24)" > .env` with a commented-out key. That flips the first-run default from authenticated to open, and the prose that replaced it is correct but skippable. I put the generating line back into the quick-start block with the condition on the same line, and made `.env.example` say that the server has no authentication at all without it. This is the one thing in the PR I would not have taken as-is: everything else is additive, and that one was a removal.

The `.wslconfig` memory/swap entry and the exit-137-during-`quant_lm_head` signature are good additions — an exit 137 with idle VRAM is exactly the kind of thing that reads as a GPU OOM and sends people to the wrong knob.

PR #78, opened 2026-09-04
docs: add optional OpenCode integration

cofade:
## Summary

- Add an optional OpenCode workspace provider for the local OpenAI-compatible vLLM endpoint.
- Document how to start OpenCode from the repository and select the served model.

## Notes

This is intentionally optional and does not change the serving stack or launch behavior. The workspace config contains no secret; the `local` API-key value is a placeholder for the default unauthenticated local endpoint.

Tested locally with OpenCode CLI 1.18.21 against `qwen3.8-27b`.


mhenrichsen:
Happy to take the documentation half of this; I would rather not take the root-level file.

**The README paragraph is useful and I will merge it.** People do point coding agents at this server and the two things they get wrong are the base URL path (`/v1`) and the served model name, which your paragraph states.

**`opencode.json` at the repository root is the part I want to avoid**, and it is about precedent rather than about OpenCode. A tool config in the root is a claim on the workspace: it is what OpenCode reads when anyone runs it anywhere in a clone of this repo, including people who have never heard of it and are here for the serving stack. The next PR is `.continue/`, then `cline.json`, then a `.vscode/settings.json`, and none of them can be tested by CI or by me. The repo has one job and it is not being a client workspace.

Two smaller things about the contents, which matter if it becomes documentation instead:

- `"context": 65536` is right only for `CTX=fast`. `CTX=long` is 131072 on the dflash2 path and `CTX=huge` is 245760, and a client that under-declares the window will truncate silently. Worth one clause.
- `"apiKey": "local"` is fine as a placeholder, but as of `ccd714b` the docker quick start generates a real `VLLM_API_KEY` again by default, so for most people following the README from the top the placeholder will be wrong rather than unnecessary. Say `$VLLM_API_KEY` and mention it can be anything when auth is off.

**Concretely**: drop `opencode.json`, put the JSON inline in the README paragraph as a fenced block with a line saying where OpenCode expects it (`opencode.json` in the directory you run it from, or `~/.config/opencode/`), fix those two clauses, and I will merge it as it stands. That way someone who wants it copies four lines, and someone who does not gets nothing they did not ask for.

Tested-with versions in the PR body are appreciated — CLI 1.18.21 against `qwen3.8-27b` is exactly the detail that makes a docs paragraph age visibly instead of silently.

cofade:
Thanks, that makes sense, and I agree with the concern. I moved the example into the README instead of keeping a repo-root config, and tightened the wording around the API key and context-window caveat so it matches the default startup behavior. It should now be much clearer that this is optional documentation only, not a workspace-level config change.

Issue #73, opened 2026-09-04
DFlash2 acceptance regression on vLLM 0.28.0: 3.19 -> 2.66 tok/step, decode 120.5 -> 101.4 tok/s (prefill and GSM8K unchanged)


mhenrichsen:
## Summary

On the reference RTX 3090, main at `cd2fa94` (vLLM 0.28.0) accepts **~16% fewer tokens per verify step** than `8d832f8` (vLLM 0.27.1) with DFlash2, and decode throughput falls by the same amount. Prefill and quality are unaffected. Found while validating the 0.28 stack against the production profile before migrating this box; production is staying on 0.27.1 because of it.

## Numbers

Same GPU, same 250 W, same models, same `bench/prefill_ab.sh` flow so the decode guard runs in an identical position (warmup, prefill rows, then the C1 cohort of `bench/prompts_real.jsonl` at 1024 output tokens). Profile is this box's production one: `CTX=fast SPEC=dflash2 PREFIX_CACHE=1 INT8_ACT=int8 INT8_LAYERS='mlp|linear_attn|self_attn' PREFILL_ATTN=int8`. Trees interleaved, two passes each.

| tree | `DFLASH_TOKENS` | tok/step | decode C1 |
|---|---|---:|---:|
| 0.27.1 (`8d832f8`) pass a | 15 | **3.19** | 120.5 tok/s |
| 0.28.0 (`cd2fa94`) pass a | 15 | **2.66** | 101.7 tok/s |
| 0.27.1 pass b | 15 | **3.19** | 120.3 tok/s |
| 0.28.0 pass b | 15 | **2.66** | 101.4 tok/s |
| 0.27.1 | 7 | 3.12 | 117.5 tok/s |
| 0.28.0 | 7 | 2.68 | 102.4 tok/s |

Reproducible to three significant figures in both directions, and present at both verify widths, so it is not a LABD-specific effect.

**Everything else is equal.** Prefill on the same two trees is identical — 1,502 / 1,499 / 1,503 / 1,504 tok/s at 51,200 tokens across the four passes — and so is capacity (KV pool 57,669 tokens, `max_model_len` 57,344) and quality (**GSM8K 200 at 96.0% on both**, thinking off, greedy). Tool calls round-trip on 0.28. The loss is entirely in acceptance: decode tok/s tracks tok/step to within a percent, which is what you expect when the step cost is unchanged and only the yield per step moves.

## Partially bisected

[...]

PR #79, opened 2026-09-05
dflash2-z-adaptive-emitted: keep the code, fix the comment that contradicts it

cpuchip:
Keeps the code in `dflash2-z-adaptive-emitted.patch` and fixes the comment that contradicts it. Follows from #73, where you said the patch should be re-justified or dropped on its own merits and that you would take a PR either way.

**The code is right and the comment is wrong, on both engines.** `v1/worker/gpu/input_batch.py` defines the counters the same way on 0.27.1 and 0.28:

```python
num_rejected = num_logits - num_sampled
```

So `num_sampled` is the count the sampler emitted, bonus plus accepted drafts, and `num_rejected` is the slots it did not fill. The comment says the opposite, that `num_sampled` counts the slots given and the difference is what was emitted. Under the true definitions the 0.27.1 formula `num_sampled - num_rejected` works out to twice the emitted count minus the slot count, which double-subtracts every rejection. The patch replaced that with `num_sampled`, which is the emitted count, so the commit title "Fix DFlash2 adaptive accepted-token accounting" was accurate and the comment above it was stale from the day the formula was first written.

This PR changes the comment to state the definitions and why the subtraction is wrong, and leaves both code hunks as they are. The regenerated patch applies cleanly to a pristine 0.28.0 tree with the rest of the series and reproduces the same source, plus five comment lines.

**On the 2.66 to 2.79 you measured from reversing it.** Reversing reintroduces the double subtraction, which feeds the block-length policy an emitted count that is too low, so the long block is entered less often. On your cohort that scored higher. That is a real effect and it is worth having, but it is a tuning result reached by feeding the policy a wrong number, and it will move the moment acceptance changes for any other reason. The band analysis from #73 says the same thing from the other side: the two formulas disagree only on steps that sample at least eight tokens and reject enough to fall back under eight, so the gain is confined to workloads that live in the long block. If shorter blocks are better for your cohort, the honest way to get them is the policy's own threshold, which is `1 + draft_block` in `next_num_draft_tokens`, exposed as a knob or raised by one. I have not measured that and would not merge a threshold change on this evidence; I am naming it so the 0.13 does not get chased through the accounting again.

Runtime-verified on the 3090 side of the earlier work: the predicate reads the corrected value on 98 of 99 steps identically under either formula on a short prompt, which is why single-prompt cells could not see this patch at all.

### DIFF
--- a/patches/dflash2-z-adaptive-emitted.patch
+++ b/patches/dflash2-z-adaptive-emitted.patch
@@ -1,21 +1,26 @@
-diff --git a/v1/worker/gpu/spec_decode/dflash/speculator.py b/v1/worker/gpu/spec_decode/dflash/speculator.py
-index 1bd78be21..0556bcbb4 100644
---- a/v1/worker/gpu/spec_decode/dflash/speculator.py
-+++ b/v1/worker/gpu/spec_decode/dflash/speculator.py
-@@ -362,7 +362,7 @@ class DFlashSpeculator(DraftModelSpeculator):
-         # sampling slots it was given (bonus + drafts), num_rejected how many of those were
-         # thrown away, so the difference is the tokens it emitted. dflash2/lookup.py decides
-         # the next block length from it.
+--- a/v1/worker/gpu/spec_decode/dflash/speculator.py	2026-09-05 00:21:38.703248441 +0000
++++ b/v1/worker/gpu/spec_decode/dflash/speculator.py	2026-09-05 00:21:38.735250379 +0000
+@@ -358,11 +358,12 @@
+         is_profile: bool = False,
+     ) -> torch.Tensor:
+         num_reqs = input_batch.num_reqs
+-        # What the step that just ran actually produced per request: num_sampled counts the
+-        # sampling slots it was given (bonus + drafts), num_rejected how many of those were
+-        # thrown away, so the difference is the tokens it emitted. dflash2/lookup.py decides
+-        # the next block length from it.
 -        self.last_num_emitted = (num_sampled - num_rejected) if num_sampled is not None else None
++        # What the step that just ran actually produced per request. num_sampled is the
++        # count the sampler EMITTED (bonus + accepted drafts): input_batch.py defines
++        # num_rejected = num_logits - num_sampled, so subtracting num_rejected here would
++        # double-count the rejections. dflash2/lookup.py decides the next block length
++        # from this number.
 +        self.last_num_emitted = num_sampled if num_sampled is not None else None
          num_target_tokens = input_batch.num_tokens
          num_query_tokens = num_reqs * self.num_query_per_req
          max_seq_len = input_batch.seq_lens_cpu_upper_bound[:num_reqs].max().item()
-diff --git a/v1/worker/gpu/spec_decode/dflash2/speculator.py b/v1/worker/gpu/spec_decode/dflash2/speculator.py
-index 1371aab30..ac0d41df0 100644
---- a/v1/worker/gpu/spec_decode/dflash2/speculator.py
-+++ b/v1/worker/gpu/spec_decode/dflash2/speculator.py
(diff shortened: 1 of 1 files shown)


cpuchip:
Measured the claim in the body rather than leaving it as a claim, and it comes out weaker than I wrote it, in a way that matters for what this patch is worth now.

Three arms on the eight-prompt cohort at 1024 output tokens, four seeds, paired by seed and prompt, `draft_sample_method` set in all three so this is the configuration production runs now. Tokens per step is the metric, degeneracy guard applied first (it removed nothing), threshold registered before the run.

| arm | tokens per step | drafted tokens per round |
|---|---|---|
| shipped patch, coast 3 | 3.735 | 7.140 |
| patch reversed, coast 3 (your arm) | 3.739 | 7.113 |
| shipped patch, `VLLM_DFLASH2_LOOKUP_STICKY=0` | 3.737 | 7.055 |

Paired against shipped: reversed is +0.004 tokens per step at t = 0.13, coast zero is +0.003 at t = 0.10. Both arms lowered long-block occupancy as predicted, and neither moved throughput. The same null holds inside each stratum, the rows where the base never left the drafter's block and the rows where it did.

So the 2.66 to 2.79 you measured from reversing this patch does not reproduce on this box with the field set. I think the reason is the regime, and it is the same reason as the original bug. You measured it with the field unset, where the drafter picks by argmax and acceptance is lower, so the emitted count sits nearer the policy's threshold and the two formulas disagree more often. With the field set, acceptance is higher, more steps clear the threshold under either formula, and the difference between them stops mattering. That reading is consistent with the per-step instrumentation from the other box, where the two formulas disagreed on one step in ninety-nine.

Which means the body's last paragraph should be read as: the honest knob exists, it does what it says to block occupancy, and on the production configuration there is nothing left for it to recover. The comment fix stands on the definitions alone. If you still see 0.13 with the field set on your cohort, that is a box difference worth knowing and I would want your drafted-tokens-per-round beside it.

mhenrichsen:
Merging. Your read of the counters is right, and the red check was mine, not yours.

**The code is right and the comment is wrong, confirmed at the source.** `v1/worker/gpu/input_batch.py:518` in the installed 0.28.0 tree:

```python
num_rejected = num_logits - num_sampled
```

So `num_sampled` is what the sampler emitted and `num_rejected` is the unfilled slots, exactly as you say, and the shipped comment states the opposite. Applying your patch to a v0.28.0 checkout with the four preceding patches gives byte-identical code to `main` plus your five comment lines — I diffed the two results with comments stripped, both files identical.

**On the failing `git-apply` check: the job was broken, and your patch is what exposed it.** `git apply` resolves a patch path against the *repository* root and silently skips anything outside the subdirectory it runs in. `check_vllm_series.sh` was invoked as `git -C .ci/vllm/vllm apply`, while the patches address files relative to the installed vllm package (`v1/worker/...`), so every path fell outside the prefix and every file was skipped. I reproduced it against a v0.28.0 checkout: all five patches "apply", the tree is clean afterwards, exit 0, `patch integrity: OK`. The job has been checking nothing.

Your patch is in traditional diff format rather than `git diff` format, which `git apply` does *not* skip — so it was the only patch in the series ever actually checked, and it was checked against bare upstream instead of against the four patches that precede it, where the line it removes does not exist yet. Hence `patch failed: ...:358`.

Fixed in `d21ed22`: run from the repository root with an explicit `--directory`, and fail if a patch that applied changed no file, so the vacuous case cannot come back quietly. With the fix the series really applies (11 files, 871 insertions) and **your patch passes** — I ran the fixed checker with your file in place of `main`'s and it goes green through all five.

Nothing for you to change. Thank you for the re-justification; the tuning question from #73 stays open on its own terms, and I agree with your framing that a gain reached by feeding the policy a wrong number is not the way to keep it.

Issue #75, opened 2026-09-04
Autotune nondeterminism decides the trajectory: a bench row needs its compile-cache state


cpuchip:
Filing this as its own issue so the compile-cache rule from the #25 tracking update has somewhere to point. It is not a defect in this repo; it is a property of the stack that makes benchmark rows incomparable unless you state one more thing about how they were produced.

## What happens

Six Triton kernels are autotuned on this model, all of them in the chunked Gated DeltaNet path. Autotuning times candidate configurations at first use and keeps the winner. vLLM caches those winners, so a restart does not re-run the race, but a fresh container or `VLLM_DISABLE_COMPILE_CACHE=1` does. A different winner means a different partition of the same tiles, so the fp32 accumulation happens in a different order, the downcast to bf16 rounds a few elements the other way, and at greedy decoding a long tool-using conversation can land on the other side of a near tie and never come back. Within one boot it is exactly deterministic; across boots without a cache it is not.

## The measurements

On my box, a twelve-turn tool-calling conversation, temperature 0, thinking off, same image and arguments every time:

| condition | runs | result |
|---|---|---|
| cache persisted, two boots | 2 | identical, 144 tool calls both times |
| no cache, fresh container each boot | 4 | 124, 189, 78, 74 tool calls |
| combo-kernel benchmarking off, no cache | 3 | 160, 77, 160 |

On a second box, native and bare rather than containerised:

| condition | runs | result |
|---|---|---|
| one boot, two conversations | 2 | identical call for call |
| cache persisted, two boots | 2 | identical, wall time within 0.1 s |
| both caches deleted before each run | 2 | identical, 160 calls both times |

That last row is the useful one. Deleting the cache on a quiet bare box does not scatter the results; it moves the fixed point and keeps it fixed. So the rule is not "no cache means nondeterminism", it is that a missing cache forces the timing race to be re-run and the race is only as reproducible as the box's timing is. Container noise is what makes it vary. Disabling the benchmarking is not sufficient, as the third row above shows.

## Which kernel

The persisted cache records every candidate's measured time, so a winner can be forced by rewriting one entry and rebooting. Three of the six kernels chose differently between a good draw and a bad one. Forced one at a time, each way, pins verified to survive the boot:

| tree | kernel forced | result |
|---|---|---|
| good | `chunk_scaled_dot_kkt_fwd_kernel` to the bad choice | unchanged |
[...]

PR #80, opened 2026-09-05
gotchas: five from the 0.28 acceptance bisect

cpuchip:
Five entries from the #73 and #75 work, in the file's own shape: the lead sentence is the thing that costs hours, the body carries the measurement that established it and the fix.

53. A benchmark row without its compile-cache state is not reproducible.
54. A high acceptance rate can mean the drafter is good or the generation has collapsed.
55. Report tokens per step, not accepted per drafted token; the denominator is order-dependent.
56. At width 15 on ordinary text the engine drafts 7 and queries 8, so 15 and 7 are one experiment.
57. Two passes with a fixed seed are two replays, and a short single-prompt cell cannot see a cohort-scale effect.

Every number in them was measured this week on either the reference 3090 or an RTX 4090 under WSL2, and the write-ups they summarise are on the two issues. I left out the "instrument the distribution, not the mean" line because you said you were adding it with attribution, so nothing here collides with that. If your entry lands first the numbers shift by one, which is a trivial conflict.

The one I would most want a second reader on is 55, because it turns on a code read of the sticky branch in `next_num_draft_tokens` plus one ordering experiment; the other four have measurements from both boxes behind them.

### DIFF
--- a/docs/gotchas.md
+++ b/docs/gotchas.md
@@ -882,3 +882,75 @@ Things that each cost us hours, in rough order of pain. Worth skimming before yo
     block is not free on int4**: at 15 the pool is 53,908 tokens against
     142,843 at 7, and the 256k default no longer fits (`estimated maximum
     model length is 180320`, a clear `ValueError` rather than an assert).
+
+53. **A benchmark row without its compile-cache state is not reproducible, because the
+    autotuner's timing race picks the kernels and the kernels pick the trajectory.**
+    Filed as [#75](https://github.com/syv-ai/qwen38-27b-rtx3090/issues/75). Six Triton
+    kernels in the chunked Gated DeltaNet path are autotuned at first use; vLLM caches the
+    winners, but a fresh container or `VLLM_DISABLE_COMPILE_CACHE=1` re-runs the race, and a
+    different winner is a different reduction order, a different last bit, and at greedy a
+    different token at the first near tie. Measured on one image with the cache wiped before
+    each boot: 8 boots, 8 distinct winner sets, 5 distinct trajectories, the same twelve-turn
+    tool conversation ranging from 58 to 189 tool calls. With the cache persisted, two boots
+    were identical to the call. So: mount a persistent cache into bench containers rather than
+    disabling it for hygiene, and copy the `*.autotune.json` records beside a published row so
+    a reader can tell whether two rows are even comparable. A quiet bare box re-times to the
+    same winners even with the cache deleted; container timing noise is what makes the draw
+    vary. The same cache state also moves the profiled peak activation and therefore the KV
+    pool, by 0.92 GiB on the reference 3090, which is a separate channel with a separate fix
+    (state it, or pin `--kv-cache-memory`).
+
+54. **A high acceptance rate can mean the drafter is good or the generation has collapsed,
+    and the counter cannot tell you which.** Degenerate text is trivially predictable. One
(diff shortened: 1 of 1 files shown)


cpuchip:
Amended entry 55 after replicating it on a second box, because the replication came out stronger and messier than the entry said.

The original claim was that request order moves drafted tokens while drafts and accepted tokens stay fixed, so tokens per step is order-invariant. That held on the quiet 3090 it was measured on. On an RTX 4090 under WSL2, with a prompt that keeps the long block engaged on every seed, the same six-seeds-two-orders design changed drafts and accepted tokens too on four of six seeds, one of them by a factor of two. So on that box tokens per step itself moves with order.

Two mechanisms are consistent with both boxes and the entry now names both: the sticky long block coasting into the next request, which changes requested block length; and a different block length being a different verify batch shape, which on a numerically knife-edged model can flip a near-tie token even with the seed fixed. A third, prefix caching, applies whenever requests share a prompt, as they did in both experiments.

The prescription is unchanged and is now the whole point of the entry: hold request order fixed within a comparison, report tokens per step rather than acceptance per drafted token, and treat per-request figures from a sequence as dependent samples. What changed is that the entry no longer promises tokens per step is invariant, because on one of two boxes it is not.

A run with prefix caching off is queued to separate the sticky mechanism from the cache one. If it changes the entry again I will say so here.

cpuchip:
The prefix-caching arm ran, and entry 55 is final.

Same six seeds, same two orders, same context-reproducing prompt, engine started with `--no-enable-prefix-caching` and the server log confirming `enable_prefix_caching=False`. The rows changed against the cache-on arm, so the switch took, and the order effect persisted: four of six seeds changed drafts and accepted tokens with order, one of them from 31 rounds to 112. So on this box the driver is the block-length policy's coast into the next request, and prefix caching is ruled out as the cause. The entry now says that rather than listing caching as a live third mechanism.

One method note, because the first version of that arm was wrong in a way worth knowing. Setting the launcher's `PREFIX_CACHE=0` adds nothing to the engine arguments; it leaves the engine default, and for a generative hybrid model `is_prefix_caching_supported` returns true, so the engine turned caching on anyway and the arm came back bit-identical to the cache-on run. A zero that adds nothing is not an off switch. The rerun passes `--no-enable-prefix-caching` through `EXTRA_ARGS` and greps its own resolved setting out of the server log before it reports a number. I would not trust any A/B on this repo that does not print the engine's resolved configuration beside its result, mine included.

mhenrichsen:
Merging as 53-57.

I can corroborate 56 from an independent source: production traffic on the reference 3090, `DFLASH_TOKENS=15`, from this evening's log. The per-position acceptance rates split exactly where you say they do:

```
Per-position acceptance rate: 0.958, 0.921, 0.899, 0.862, 0.845, 0.820, 0.793,
                              0.155, 0.155, 0.155, 0.155, 0.155, 0.155, 0.153, 0.148
```

Seven positions in the 0.79-0.96 band and eight in a flat 0.15 tail, over four consecutive metrics windows under real load. That is the clamp to the trained block showing up from the other side, and it is a cleaner tell than the drafted-per-round histogram because it needs no instrumentation — it is in every log anyone already has. I have added that sentence to 56 rather than a sixth entry.

53 and 55 line up with what I measured from the memory side on #75 (the 0.92 GiB compile-cache swing, and the position-in-run channel that gave me a wrong-signed 0.28-versus-0.27.1 decode result). 57 is the one I most needed written down, because the underpowered short cell was mine.

Thank you for testing prefix caching on 55 rather than leaving it as the obvious candidate.

PR #81, opened 2026-09-05
single-user: width-15 default context 56320, off the KV-cache edge

cpuchip:
1. `DFLASH_TOKENS=15` defaulted to a 57344-token context, which sits about 10 MiB from the pinned pool's edge: a 3090 refuses the boot (`estimated maximum model length 57120`), and a 4090 under WSL2 boots it only because the shortfall spills into host memory and runs slow instead (issue #25, item 9 and its correction).
2. 56320 (55 x 1024) is one thousand tokens less and about 78 MB of pool at this geometry, so the boot is deterministic on both cards at the shipped pin. Nothing else changes; `DFLASH_MAX_LEN` still overrides.

The launcher comment carries the numbers. Measured on an RTX 3090 and two RTX 4090s under WSL2; the `#25` comments have the boots.

### DIFF
--- a/single-user/start_qwen.sh
+++ b/single-user/start_qwen.sh
@@ -354,11 +354,15 @@ if [ "$SPEC" = "dflash2" ]; then
       export VLLM_V2_CUDAGRAPH_MEM_MIB=${VLLM_V2_CUDAGRAPH_MEM_MIB:-1400}
     fi
   elif [ "$DRAFT_TOKENS" -gt 7 ]; then
-    # 4 slots and 56k instead of 8 and 64k: the aligned state pages and the bigger decode
+    # 4 slots and 55k instead of 8 and 64k: the aligned state pages and the bigger decode
     # graphs are what the long block costs, and this is where they still fit next to the
-    # 5.2 GiB pool (57,669 tokens). DFLASH_TOKENS=7 gets 8 slots and 64k back.
+    # 5.2 GiB pool. DFLASH_TOKENS=7 gets 8 slots and 64k back. 56320 rather than 57344:
+    # at this pin one 57344-token request needs about 10 MiB more than the pool holds, so a
+    # 3090 refuses the boot ("estimated maximum model length 57120") and a 4090 under WSL2
+    # boots only because the shortfall spills into host memory (#25, item 9). 1024 tokens
+    # is about 78 MB of pool at this geometry, so 55 x 1024 leaves real margin.
     MAX_SEQS=${MAX_SEQS:-4}
-    MAX_LEN=${DFLASH_MAX_LEN:-57344}
+    MAX_LEN=${DFLASH_MAX_LEN:-56320}
     KV_MEM=${KV_MEM-5583457484}
     # Decode graphs are captured for both block lengths (the drafter's and the full verify
     # block), or the short step -- the common one -- runs piecewise and costs 8%. That is


mhenrichsen:
Holding this one on a question, and I want to be precise about which part I am questioning, because the margin argument and the refusal claim are separable.

**The refusal does not reproduce on the reference 3090.** Cold boot, GPU idle at 18 MiB beforehand, repo defaults with no drop-in, `SPEC=dflash2 CTX=fast PREFIX_CACHE=1 DFLASH_TOKENS=15`, straight off `main`:

| arm | pool | request | headroom |
|---|---:|---:|---:|
| `main` (57344) | 57,669 tokens | 57,344 | **325** |
| this PR (56320) | 57,291 tokens | 56,320 | **971** |

Both boot to health. The `main` arm has also been serving production on this card since I restarted it this evening, at the same 57,669 / 57,344 / 1.01x, and it ran for a day and a half at those numbers before that. So I cannot get `estimated maximum model length 57120` out of a 3090 here, and 57,120 against my 57,291-57,669 is a real difference in *pool*, not in the request — which is odd, because `KV_MEM` is pinned on this path, so the pool ought to be a function of geometry alone.

**What I would like before changing a shipped default**: the boot log from the 3090 that refused — the `GPU KV cache size` line if it got that far, the `non-default args` line, and whether that box was WSL2 or bare metal. If its pool really is 57,120 at the same pin, then something in the geometry differs between our trees and that is worth more than this one-line change; if it was the WSL2 3090, then #83's ceiling is the mechanism and this default is being set by a host-memory limit rather than by the pool, which I would want the comment to say.

**The margin argument stands on its own and I am sympathetic to it.** 325 tokens is 0.57% and the pool is not independent of the request length — my own two arms above differ by 378 tokens of pool from nothing but `MAX_LEN` moving. Anything that shifts the state pages (a drafter change, a vLLM point release, a patch landing) can cross a gap that thin. I would take the change on that argument alone, with the comment rewritten to say what is measured rather than that a 3090 refuses.

So: post the log and I will merge it as written, or say the word and I will merge it with the comment reworded to the margin argument and your box's refusal attributed to WSL2. Either way it needs a one-line rebase now — #82 landed first and moved that same line to `MAX_LEN=${DFLASH_MAX_LEN:-${USER_MAX_LEN:-57344}}`.

cpuchip:
You are right on every count, and I owe you the plain version.

The refusing boot's log does not exist. That box overwrites its server log on every boot and the failed cell was never archived, so the only place 57120 lives is my own write-up. Rather than defend it I tried to reproduce it. Seventeen cold boots on the same 3090 (bare metal; the WSL2 qualifier in the body belongs to the two 4090s only), from `main` at 7ca84fc at width 15: twelve at the 57344 default across persisted and fresh compile caches, three with 311 MiB and 1 GiB deliberately held on the card by another process, and two with `SPEC_ATTN=0`, which is the knob the September 5 cell actually ran. Every boot at 57344 came up with `GPU KV cache size: 57,669 tokens`. The one with 1 GiB held built that same pool and then died in graph capture with a CUDA out-of-memory, a different failure from the ValueError. Nothing reproduced the refusal. The claim that a 3090 refuses this default is withdrawn.

The boots also explain your range, which turns out not to be one. With `kv_cache_memory_bytes` set, the engine prints that it skipped memory profiling, so the pool is a function of the pin and the geometry alone and moves with nothing else: 57,669 tokens at 57344 and 57,291 at 56320. Those two numbers are the before and after of this change. Your instinct that the pool ought to be a function of geometry alone is right, and our seventeen boots agree with yours to the token.

So this PR no longer has a rationale of its own. Its claim does not reproduce, and I am not going to ask you to move a shipped default on my say-so when the only argument left is one you made better than I did. I am closing it as cannot reproduce. I am sorry for the evidence I lost when that boot's log went unarchived, and for the evening it cost you chasing a sentence I could not produce. If the margin question comes back with a measurement behind it, it will come back as a new PR that carries the measurement.

PR #82, opened 2026-09-05
single-user: honour MAX_LEN on the DFlash path; warn when the drafter outgrows the pinned budgets

cpuchip:
1. `MAX_LEN` is honoured on the `SPEC=dflash2` path. The five profile blocks reset it from `DFLASH_MAX_LEN`'s default, so `MAX_LEN=8192 SPEC=dflash2 ...` booted at 65536 and said so only in the engine's args line. Precedence is now `DFLASH_MAX_LEN` > `MAX_LEN` > the profile default, and the profile defaults are unchanged.
2. A warning when the drafter's weights outgrow the pinned budgets. The KV pin, `GPU_UTIL` and the graph reservation are sized for the shipped 1.2 GiB head, none of them knows the drafter's size, and none of the engine's error messages names them. A drafter over 2 GiB at the pinned defaults now gets one message naming the three budgets and the two profiles measured to serve a 4 GiB bf16 drafter on 24 GiB (`KV_MEM=3000000000 DFLASH_MAX_LEN=8192`, or `GPU_UTIL=0.97 KV_MEM= DFLASH_MAX_LEN=8192`); with either variable set it prints a one-line note instead. A warning, not a refusal: a bigger card has room where 24 GiB does not.

Both from issue #25, items 13 and 14. Checked in the container image on an RTX 4090: `MAX_LEN=8192` resolves to 8192 (it resolved to 65536 before); the 4 GiB community drafter at the defaults prints the warning; the same drafter with the fitting profile prints the note and resolves 8192 with the 3 GiB pin; the defaults with the shipped head print neither and resolve 65536.

### DIFF
--- a/single-user/start_qwen.sh
+++ b/single-user/start_qwen.sh
@@ -146,6 +146,12 @@ CTX=${CTX:-fast}
 SPEC=${SPEC:-mtp}
 # SPEC_ATTN=1: split-KV Triton attention for the multi-query verify step
 # (patches/spec-decode-attn.patch); bf16 KV only, so CTX=fast only.
+# Remember whether MAX_LEN came from the caller: the SPEC=dflash2 profiles below pick
+# their own context default per profile, and used to overwrite a caller's MAX_LEN with
+# it, so `MAX_LEN=8192 SPEC=dflash2 ...` booted at 65536 and said so only in the
+# engine's args line (#25, item 13). Precedence on that path is now
+# DFLASH_MAX_LEN > MAX_LEN > the profile default.
+USER_MAX_LEN=${MAX_LEN:-}
 if [ "$CTX" = "fast" ]; then
   MAX_LEN=${MAX_LEN:-65536}
   DRAFT_TOKENS=${DRAFT_TOKENS:-4}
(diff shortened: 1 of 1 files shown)


mhenrichsen:
Merging. Both halves verified on the reference 3090 against a launcher whose `vllm` is a stub that prints its argv, so the resolution is read rather than inferred.

**MAX_LEN precedence**, twelve configurations, `main` against this branch, full argv diffed:

| config | main | this PR |
|---|---:|---:|
| `SPEC=dflash2 MAX_LEN=8192` | 65536 | **8192** |
| `SPEC=dflash2 DFLASH_TOKENS=15 MAX_LEN=8192` | 57344 | **8192** |
| `SPEC=dflash2 CTX=long MAX_LEN=8192` | 131072 | **8192** |
| `SPEC=dflash2 CTX=huge MAX_LEN=8192` | 245760 | **8192** |
| `DFLASH_TOKENS=15 DFLASH_MAX_LEN=40000` | 40000 | 40000 |
| the other seven, incl. every unset-MAX_LEN default and all of `SPEC=mtp` | — | **argv byte-identical** |

So `DFLASH_MAX_LEN` still wins, the profile defaults are untouched, and nothing outside the dflash2 path moves.

**The drafter-size warning**, six branches, using the repo's own 3.6 GiB bf16 `Qwen3.8-27B-DFlash2` as the oversized case:

- shipped 1.2 GiB head, defaults: silent (both `main` and this branch)
- 3.6 GiB drafter, defaults: the WARNING, naming `KV_MEM=5583457484 pinned, GPU_UTIL=0.93, VLLM_V2_CUDAGRAPH_MEM_MIB=1400, MAX_LEN=65536`
- 3.6 GiB drafter + `KV_MEM=3000000000 DFLASH_MAX_LEN=8192`: the note
- 3.6 GiB drafter + `GPU_UTIL=0.97 KV_MEM= DFLASH_MAX_LEN=8192`: the note, correctly reading `KV_MEM=unpinned`
- 3.6 GiB drafter + `MAX_LEN=8192` alone: the note

Naming the three budgets in the message is the right call — none of them appears in any error the engine produces, which is exactly why item 13 took seven boots.

One thing I did **not** change and want on the record: this now conflicts textually with #81, which moves the same width-15 default line. #81 is held on a separate question (see my comment there), so this went in first and #81 will need the one-line rebase, not the other way round.

PR #83, opened 2026-09-05
docs: gotcha 53, the WSL2 dedicated-memory ceiling

cpuchip:
1. Gotcha 53: on WSL2 the card's usable dedicated memory is about half a gigabyte smaller than the same card on bare metal (23,371 MiB against 23,859 free at boot on a 3090), the shipped `SPEC=dflash2` boot sits about 50 MiB under that line, and anything larger runs two to six times slower instead of failing, with nothing in vLLM's log to show it. Four "4090 costs" reported in #25 over the past two days were this one thing (width 15, a query maximum of 16, a mixed compile, a bf16 drafter at the shipped pin); the table has each over the line and with room.
2. What to do about it: read the Windows adapter-memory counters rather than the log (shared usage at its idle baseline means the working set is on the card; gotcha 43's `dmon` signature is the same state from the other side); trust acceptance columns from a spilled run and not its tok/s; and give it room with `KV_MEM=3000000000 DFLASH_MAX_LEN=8192`, which costs the shipped head nothing at width 7, or `SPEC_ATTN=0`, which frees about a gigabyte of capture on either host.
3. One paragraph in the README's WSL2 section pointing at it.

Measured on two RTX 4090s under WSL2 with a bare-metal 3090 as the control; the boots are in the #25 comments (items 13 and 14 and the corrections to items 9 and 11). Written beside #77 rather than into it; it touches the same README section, so whichever merges second is a one-paragraph rebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,15 @@ name: `VLLM_WSL_PIN_MEMORY` is **not** a vLLM variable and setting it does
 nothing — this README named it for 22 minutes on 2026-08-21 (`589daae`, fixed in
 `27f51fa`), so a tree cloned in that window will have it.
 
+**On WSL2 the usable dedicated memory is about half a gigabyte less than the
+same card on bare metal, and the shipped `SPEC=dflash2` boot sits about 50 MiB
+under it.** Anything larger (a wider verify block, a bigger drafter, a raised
+pin, a boot that recompiles a graph) runs two to six times slower instead of
+failing, and the log does not say so; `nvidia-smi` looks the same either way.
+Gotcha 53 has the counters to read, the four costs that turned out to be this,
+and the profile that gives it room (`KV_MEM=3000000000 DFLASH_MAX_LEN=8192`,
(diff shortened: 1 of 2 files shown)


mhenrichsen:
Merged as **gotcha 58**, not 53 — #80 landed 53-57 a few minutes earlier, so I did the one-paragraph rebase you offered rather than send it back. Content is unchanged; the README pointer moved with the number.

The part I want to single out is the third bullet: *boot class and memory state have to be matched before two hosts are compared*. That is the one that retires a whole class of cross-host arguments in the #25 thread, including some I helped keep alive. Four separately reported "4090 costs" collapsing into one ceiling is a better result than four fixes would have been.

The `--gpus device=N` finding is a genuinely nasty one and I am glad it is written down beside the rest — a device request that is recorded and not enforced is the kind of thing that silently invalidates every row in a two-card comparison.

Nothing outstanding on this one.

PR #84, opened 2026-09-05
dspark: loader patch for a bf16 DSpark drafter beside the quantized target; DRAFT_METHOD; README

cpuchip:
1. `patches/dspark-draft-quant-config.patch`: one hunk in vLLM 0.28.0's `model_executor/models/utils.py`. The loader asks for the draft's quantization config through the target's `hf_overrides`, which `SpeculativeConfig` composes as a callable for the DFlash family, and `get_quant_config` refuses a callable, so a bf16 DSpark drafter beside the W4A16 target dies at load. The draft carries no `quantization_config`, so the patch returns None in that case. Header in the style of the other patches; `_check_applied.py` sees it.
2. `DRAFT_METHOD` in the launcher (default `dflash`, unchanged behaviour), so running a DSpark drafter is an environment variable rather than an edit of the `SPEC_CFG` line.
3. A README paragraph on what a published Qwen3 DSpark checkpoint needs: this patch, and a copy whose `config.json` names `Qwen3DSparkModel` (the registry maps the published architecture name `DSparkDraftModel` to the DeepSeek V4 class, which is upstream's to fix, not a patch here). It says the drafter loses to the shipped head on the same requests, so it is documentation, not a recommendation.

Checked in the container image on an RTX 4090: the patch dry-runs and applies to the installed tree, `patches/_check_applied.py` reports it present, and the launcher with `DRAFT=<renamed copy> DRAFT_METHOD=dspark KV_MEM=3000000000 DFLASH_MAX_LEN=8192` boots to health in 270 s with `method: dspark` in the engine's args, serves a coherent 120-token completion, and reports 36 drafts, 252 draft tokens, 89 accepted on it. The measurements behind the README's numbers are in issue #25 (items 15 and 16).

Wording: "tokens per step" in the README paragraph counts accepted drafts plus the bonus token, the usual speculative-decoding figure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,20 @@ name: `VLLM_WSL_PIN_MEMORY` is **not** a vLLM variable and setting it does
 nothing — this README named it for 22 minutes on 2026-08-21 (`589daae`, fixed in
 `27f51fa`), so a tree cloned in that window will have it.
 
+**Running a DSpark drafter.** vLLM 0.28.0 can serve RadixArk/Qwen3.8-27B-DSpark
+(bf16, seven drafts per step like the shipped head) once two things are in
+place: `patches/dspark-draft-quant-config.patch` (the loader refuses a bf16
+drafter beside the quantized target without it), and a copy of the checkpoint
+whose `config.json` names the architecture `Qwen3DSparkModel` instead of
+`DSparkDraftModel` (the registry maps the published name to the DeepSeek V4
+class; the weights are unchanged, so hard-link the safetensors). Then
+`DRAFT=/path/to/that/copy DRAFT_METHOD=dspark KV_MEM=3000000000
+DFLASH_MAX_LEN=8192 SPEC=dflash2 CTX=fast bash single-user/start_qwen.sh`. It
+serves, and loses to the shipped head on the same requests: 3.19 against 3.77
+accepted tokens per step and 115 against 160 tok/s on a 4090, 3.37 against 3.83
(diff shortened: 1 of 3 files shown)


mhenrichsen:
Merged, with the README conflict against #83 resolved by keeping both paragraphs (they sit in the same section).

**What I checked on the reference 3090**, since the risk here is not the DSpark path but the shipped one:

- The patch dry-runs, applies and reverses cleanly against the installed 0.28.0 tree.
- The guard is a no-op for the shipped drafter, and I confirmed the premise rather than assuming it: `Qwen3.8-27B-DFlash2-W4A16/config.json` **does** carry a `quantization_config`, so it keeps the old path; the bf16 `Qwen3.8-27B-DFlash2` carries none, which is the case that used to die.
- With the patch applied, `SPEC=dflash2 CTX=fast` boots to health and serves, pool 68,605 tokens — the unchanged default. Then I reversed the patch and left the tree as it was.

I did not run the DSpark drafter itself; that would need the renamed checkpoint copy, and your boot log on #25 plus the counters you posted (36 drafts, 252 draft tokens, 89 accepted) is the evidence for it. The README says it loses to the shipped head, which is the honest framing and the reason this is worth having written down at all.

The registry mapping (`DSparkDraftModel` → the DeepSeek V4 class) being upstream's to fix rather than ours is the right call — a patch that renames an architecture in the registry would be a landmine on the next bump.

PR #85, opened 2026-09-06
Backport vLLM SSE keep-alive + keep the stream alive during long pref…

grevinden:
…ills

Backport upstream vLLM PR #51034 (commit 585bb07c) to vLLM 0.28.0:
- add vllm/entrypoints/openai/sse_keep_alive.py
- wrap chat/completion response generators so an SSE comment ': keep-alive' is emitted while no data is being sent
- add --sse-keep-alive-interval <int> to BaseFrontendArgs (0 = off)

Used to avoid idle-stream disconnects from proxied clients (e.g. Bifrost 120s idle timeout) during long prefills or pauses between tokens. single-user/start_qwen.sh now passes --sse-keep-alive-interval 1.


mhenrichsen:
Tested this end to end on the reference 3090 and **the backport works exactly as described** — thank you, this is a real problem and the fix is the right one.

`patches/sse-keep-alive.patch` applies and reverses cleanly against the installed 0.28.0 tree, the server boots to health with `sse_keep_alive_interval` in its args, and on a streamed request with a long prefill:

```
keep-alive comment lines: 18 (first at 1.0s), data lines: 4, total 18.2s
```

So the stream is held open through an 18-second request that would otherwise have sent nothing for most of it. That is the behaviour a 120 s proxy idle timeout needs.

Two things before I merge, one small and one about scope.

**1. The interval wants a knob and a quieter default.** `--sse-keep-alive-interval 1` is hardcoded into the exec line, so every user of this repo gets a comment line every second on every stream, forever, whether or not there is a proxy in front. That is 18 extra lines on the request above and it is not free for clients that log raw SSE. The problem you are solving is a 120 s timeout; a 30 s interval has a 4x margin against it and is a hundredth of the traffic. Could you make it

```bash
SSE_KEEP_ALIVE=${SSE_KEEP_ALIVE:-30}
...
  ${SSE_KEEP_ALIVE:+--sse-keep-alive-interval $SSE_KEEP_ALIVE} \
```

or similar — a default that helps the proxied case, and `SSE_KEEP_ALIVE=0` to turn it off. I do not mind which default you pick as long as the number is reachable without editing the launcher, and the comment says which timeout it is sized against.

**2. The chat-template hardening is a separate change and I would like it as a separate PR.** `prepare/harden_chat_template.py` plus the `docker/prepare.sh` call is a good fix for a real client bug (the JetBrains array-arguments case is well described and the `is mapping` branch is the right shape), but it has nothing to do with SSE, it rewrites files inside every prepared model directory, and it deserves its own review and its own line in the history. Two specific things I would raise on it there: `python prepare/harden_chat_template.py` uses whatever `python` is on `PATH` rather than `venv/bin/python`, and an unguarded call in `prepare.sh` makes a template rewrite able to fail the whole prepare step.

Split those out and fix the interval knob and I will merge the SSE half straight away — the measurement above is already done, so it will not need another round of testing.

persano:
Verified on WSL2 / RTX 3090, same local image as above. Boot log shows `sse_keep_alive_interval: 1` in the non-default args. A cold 90K prompt (`stream=true`, prefill 105 s) delivered 101 keep-alive frames and completed with `finish_reason: stop` and `[DONE]`. This is the case agent clients hit with long contexts, and it's the reason we took the PR early.

mhenrichsen:
@persano thank you — 101 keep-alive frames through a 105 s cold prefill on a 90K prompt, on a third box, is the case this was written for and it now has two independent confirmations. Nobody is arguing about whether the backport works.

@grevinden this has not moved since `e00760a0` on 6 September, and the two things I asked for are still open. Restating them so they are unambiguous, because neither is a disagreement about the change:

1. **The interval needs to be reachable without editing the launcher.** `--sse-keep-alive-interval 1` is hardcoded into the exec line. `SSE_KEEP_ALIVE=${SSE_KEEP_ALIVE:-30}` and `${SSE_KEEP_ALIVE:+--sse-keep-alive-interval $SSE_KEEP_ALIVE}` is the whole change. Pick whatever default you want; persano's 105 s prefill clears a 120 s timeout at 30 s with three frames to spare, so the number is not the argument. I just do not want every user of this repo emitting a comment line per second forever with no way to turn it down.
2. **The chat-template hardening belongs in its own PR.** `prepare/harden_chat_template.py` and the `docker/prepare.sh` call are a good fix for a real client bug, and they have nothing to do with SSE. They rewrite files inside every prepared model directory, which deserves its own review and its own line in the history. Two things I would raise there: it runs `python` from `PATH` rather than `venv/bin/python`, and the call in `prepare.sh` is unguarded, so a template rewrite can fail the whole prepare step.

If I do not hear from you by **Thursday 18 September** I will do the split myself — take the SSE half onto a branch with the knob, credit you as the author of the commit, and open the template hardening as a separate PR under your name for review. I would rather you did it, since it is your work and you know why you chose 1 second, but this should not sit for another fortnight when the blocking piece is four lines of shell.

Nothing needs re-testing after the split. My 3090 measurement above and persano's 90K run both stand on the patch, which is not changing.

mhenrichsen:
Still holding, and still on the same two asks — nothing new is wrong with the backport. Both of them are in your court; the deadline I named (Thursday 18 September) stands.

One thing I said I would check before merging, now done. Your new-file hunk carries a Russian-localised no-newline marker:

```
170:\ В конце файла нет новой строки
```

instead of `\ No newline at end of file`. Both tools key on the leading backslash, so it is cosmetic — but `patches/check_vllm_series.sh` is this repo's authority on "the series still applies", it has never run with your patch in the glob, and the last time a patch-integrity job was assumed rather than run it turned out to be checking five files out of thirty. So I merged #85 into current main locally and ran it against a pristine v0.28.0 checkout:

```
$ bash patches/check_vllm_series.sh <pristine-vllm-v0.28.0>/vllm
== pass 1: the whole series, GNU patch, glob order
   32 patches applied, 0 with an offset
patch integrity: OK
```

`sse_keep_alive.py` is created with no trailing newline, byte-correct, and `git apply --check --directory=vllm -p1 patches/sse-keep-alive.patch` is clean too. So the marker is a non-issue and the series is fine with your patch in it. Leave it as it is.

On the split PR, two corrections/additions to what I wrote last time, because I went and read the code properly:

- The `python`-from-`PATH` point is **not** a problem in the Docker path — `docker/prepare.sh:11` already does `export PATH=/app/venv/bin:$PATH` before your call, so it is venv python there. It only holds for the bare-clone/native README path, where nothing sets that PATH. My bad for stating it flatly.
- The unguarded call is worse than untidy, and this is the part I would actually fix. `harden_chat_template.py` `main()` sets `rc = 1` whenever a template does not match its verbatim OLD block (`res == "unknown"`), and `docker/prepare.sh` runs under `set -e` with your call placed *before* the final `state`/`LEFT` check. So the day upstream revises the Qwen3 template, `prepare` aborts with a non-zero exit after the ~19.5 GB download and the requantisation have already succeeded, and the user sees a failed prepare with a ready model dir. Either `|| true` the call, or make an unrecognised template a warning and keep `rc = 1` for the `OSError` case only.

Also owed on the SSE half whenever it lands, and cheap: a row for the knob in the `single-user/README.md` env table, and a line in `docs/gotchas.md` — there is currently nothing in there about SSE, proxies or idle timeouts, and "every stream emits a comment line by default" is exactly what that table exists to tell people.

To restate the blocking asks so there is no ambiguity:

1. Make the interval reachable without editing the launcher. `SSE_KEEP_ALIVE=${SSE_KEEP_ALIVE:-30}` plus `${SSE_KEEP_ALIVE:+--sse-keep-alive-interval $SSE_KEEP_ALIVE}` on the exec line, with a comment naming the timeout it is sized against. Right now it is `start_qwen.sh:728`, hardcoded at 1. If you have a reason for 1 s specifically — something that bites at 30 s that neither my 18.2 s run nor persano's 105 s prefill would have shown — say so and I will take 1 as the default.
2. Split `prepare/harden_chat_template.py` and the `docker/prepare.sh` call into their own PR.

If I do not hear from you by 18 September I will do the split myself, with the knob, keeping you as the author of the commit, and open the template hardening separately under your name. No re-testing is owed either way: the patch is not changing, my 18-keep-alive-line measurement and persano's 101 frames both stand, and the series check above now covers it.

mhenrichsen:
Done — I did the split myself, three days early. Closing this in favour of two PRs, both with your commits and your authorship intact.

**#115 is merged.** That is `7e3aa04f` (the patch) and `167e1bf6` (the series entry) untouched, plus one commit of mine doing the four lines of shell I kept asking for:

```bash
SSE_KEEP_ALIVE=${SSE_KEEP_ALIVE-30}   # no colon: SSE_KEEP_ALIVE= keeps the empty value
...
  ${SSE_KEEP_ALIVE:+--sse-keep-alive-interval $SSE_KEEP_ALIVE}
```

30 s is the default, sized against the 120 s Bifrost timeout with a 4x margin — persano's 105 s cold prefill clears it with three frames to spare. `SSE_KEEP_ALIVE=0` passes the value vLLM reads as off, and an *empty* `SSE_KEEP_ALIVE=` drops the flag entirely, which matters for a second reason nobody stated here: the flag only exists because `patches/sse-keep-alive.patch` is applied, so an empty value keeps the launcher bootable against a vLLM tree that has not had the series run over it, and the deploy tree does drift from the repo. Hence `${VAR-30}` rather than `${VAR:-30}`; with the colon an explicit empty value silently becomes 30. Also landed the two doc lines that were owed: the `SSE_KEEP_ALIVE` row in the single-user README env table, and gotcha 55, which is where "every stream carries a comment line by default" belongs.

I re-ran the series check on the merged branch against a pristine v0.28.0 checkout, since the patch file and `patches/series` disagreeing is exactly how main went red on 13 September:

```
$ bash patches/check_vllm_series.sh <pristine-vllm-v0.28.0>/vllm
== pass 1: the whole series, GNU patch, patches/series order
   36 patches applied, 23 with an offset
patch integrity: OK
```

`main` alone on the same checkout gives 35 applied with the same 23 offsets, so your patch adds none of them; `git apply --check --directory=vllm -p1 patches/sse-keep-alive.patch` is clean on the pristine tree too. CI's `patch integrity / git-apply` job is green on the merge. Nothing was re-measured, because nothing needed to be: the patch is byte-identical to what I ran 18 keep-alive lines through an 18.2 s request with, and what @persano ran 101 frames through a 105 s cold prefill on 90K tokens with. Two boxes, and the backport was never what was in dispute.

**#116 is the template hardening**, open under your name: `bd716f27` plus the `HARDEN_TEMPLATES` guard cherry-picked out of `c063cd12`. One blocking ask on it, and it is smaller than it sounds — make an unrecognised template a warning and keep `rc = 1` for the `OSError` branch. `HARDEN_TEMPLATES=0` is a real improvement but it does not cover the failure, because nobody sets it before the failure they have not had yet: with the default at 1, the day upstream revises the Qwen3 template, `main()` returns 1 on `"unknown"`, the call is before the `LEFT` check under `set -e`, and prepare exits non-zero after the ~19.5 GB download and the requantisation have already succeeded. Your own new comment describes exactly this, so I think we agree on the mechanism.

Three things did not come across, and I want to be explicit rather than have them turn up again:

- `.env.fast` and `.env.long` (`a8c4bcf3`) are dropped. `.gitignore` has `.env`, `docker-compose.yml:8` says knobs go in a `.env` next to the file and are never committed, and `.env.example` is the sanctioned template. `.env.long` also sets `MAX_LEN=170000` under its own comment saying that value needs a 5.94 GiB pool and OOM'd at 23 MiB, so anyone who copied it would inherit a config the file documents as broken. If the tuning in there is worth having, it is a commented block in `.env.example` and I will take that PR.
- The `docker-compose.yml` bind-mount of `./docker/prepare.sh` over `/app/docker/prepare.sh:ro` is not in either half. It silently pins every compose user's prepare step to whatever is in their checkout instead of the image they built. That may well be what you want on your own box, but it is a behaviour change for everyone and it needs its own PR and its own argument.
- `NO_DOWNLOAD=1` and the `BASE_MODEL_DIR`/`MODEL` fallback are unrelated to both halves. I have no objection to either — send them separately and they can go in quickly.

And one retraction of mine: the `python`-from-`PATH` point does not apply to the Docker path, `docker/prepare.sh:11` exports `/app/venv/bin` first. It only holds for the bare-clone native README path. My error.

The one question I still have for you is the one I said I would not block on: is there a reason for 1 s specifically — something that bites at 30 that neither an 18.2 s request nor a 105 s prefill would show? If there is, say so on #115 and I will change the default; it is a one-line follow-up.

Thanks for the backport. It is a real bug, the upstream fix was the right thing to reach for, and the patch was clean enough that two people reproduced it on two boxes without touching it.

PR #89, opened 2026-09-08
Backport vLLM #54282: decouple the draft's Gumbel noise from the target's

cpuchip:
Backport of vLLM #54282 (fe755c889): the draft's Gumbel noise stream is salted (offset 1 << 30) so it is independent of the target's, and each speculator states its own "position before the sampled token" convention instead of relying on the +1 inside `sample_draft`. One new file, `patches/vllm-pr54282-draft-gumbel-salt.patch`, sorted after the repo's patches; it touches `sample/gumbel.py`, `sample/sampler.py`, `spec_decode/rejection_sampler_utils.py`, `spec_decode/speculator.py` and the five speculators.

Why it matters here: it only acts under `draft_sample_method=probabilistic`, which is what this launcher ships. Before the patch the noise that re-samples a rejected draft token comes from the same Philox offset that produced the draft token, so the verifier's draw is correlated with the draft's and the accepted stream leans toward the drafter. Upstream's `test_gumbel_drafted_rejection_sample_is_unbiased` measures it: with the draft drawn on the shared stream, the way this tree draws it today, the accepted distribution fails the chi-square check at position 0 with chi2 1584.9 against a threshold of 69.8; with the patch it passes.

What I measured, RTX 4090 under WSL2, the shipped default (`SPEC=dflash2 CTX=fast DFLASH_TOKENS=7 PREFIX_CACHE=1`, int8 activations and prefill attention), fresh compile cache, the patch applied to the installed tree before the launcher ran:

1. `verify.sh --no-server` passes with it applied: `verify: OK (0 failures)`. The DFlash2 candidate-selector hunk did not apply against this tree because `dflash2-lookup-drafting.patch` moved its context, so that one hunk is ported by hand (IS_DRAFTING=True and the P-1 position, the same as the other sites); the other 35 apply as upstream wrote them.
2. Upstream's tests in the shipped image with pytest installed into a throwaway container: `test_rejection_sampler_utils.py` 44 passed with the patch. One honest note on the red: the shipped test's draft helper passes `is_drafting=True`, which the old `gumbel_sample` does not accept, so unpatched it dies with a TypeError before measuring anything. To see the bias itself I ran the same test with that one keyword removed from the helper (the test's own docstring describes this adaptation) against the unpatched tree: both cases fail the unbiasedness assertion (chi2 1584.9, df 15, threshold 69.8 at position 0). That is the red; the shipped test on the patched tree is the green. `test_gpu_gumbel_sample.py` 18 of 20 pass patched; the two failures, `test_logits_cache_columns_stay_separate_across_steps[1]` and `test_logits_cache_narrower_than_logits_is_rejected`, both test vLLM #53017 (a draft-logits-cache column stride fix from August 20 that neither 0.28.0 nor this tree carries). That bug needs a cache row wider than the logits, which this drafter does not have (the DFlash2 path caches through its own strided kernel and the draft vocabulary equals the target's), so I left it out of this PR.
3. Pool unchanged at 68,605 tokens; lookup-augmented drafting on at k=7 as before.
4. Paired cohort against the main baseline I measured yesterday (same client, the repo's eight `bench/prompts_real.jsonl` prompts, 1024 tokens, temperature 1.0, seeds 1-8, rows paired by seed and prompt):

| | main 7ca84fc | with patch | delta | paired t |
|---|---|---|---|---|
| tok/step, seeds 1-4 (n=32) | 3.730 | 3.788 | +0.057 | +0.57 |
| tok/step, seeds 5-8 (n=31) | 3.840 | 3.817 | -0.023 | -0.31 |
| tok/step, pooled (n=63) | 3.784 | 3.802 | +0.018 | +0.28 |
| accepted per drafted token, pooled | 0.3629 | 0.3646 | | |
| tok/s, pooled | 160.3 | 162.3 | +2.0 | +0.72 |

On this card acceptance does not move, which is what upstream expected (their independent trials put it at 0.1068 against 0.1071). On the native 3090 (threadchip, same client, same prompts and seeds, k=7, pool 68,605 both arms) it moved down: tok/step 3.878 to 3.757 (delta -0.121, sd 0.56, n=64, paired t -1.72), accepted per drafted token 0.3781 to 0.3612, tok/s 153.8 to 148.8, with the distinct-token and gzip ratios slightly up. Pooling the paired differences from both cards gives -0.052 tok/step (n=127, t -1.10): not resolved at this size, and the 3090's sign is the one you would expect from removing a bias toward the drafter, so treat a small acceptance cost as possible rather than excluded. One caveat on the 3090 row: that box runs a native tree (cb5679d) on which `verify.sh` reports `dflash2-lookup-drafting.patch` and `dflash2-prewarm.patch` as not applied, before and after this patch alike, so its two arms pair cleanly against each other but not like-for-like against the image tree. The row that is missing from the second seed set is seed 7 on prompt 5, where the patched run emitted a single token and stopped, so there was no speculation to count. That is a property of the prompt, not the patch: the unpatched record has the same prompt stopping after one or two tokens on seeds 2, 3 and 4 in twelve earlier arms.

What I did not measure: output quality. The bias this removes is in the distribution of accepted tokens, and a cohort's tok/step cannot see that; the unbiasedness test is the oracle for it. The distinct-token and gzip ratios the client reports did not change (0.543 to 0.545 and 0.451 to 0.443).

With all three backports applied together (this, #54373 and #54374) the same boot serves the same pool and the cohort is indistinguishable from this patch alone (tok/step 3.788 against 3.759 on seeds 1-4, paired t -0.33; 3.817 against 3.844 on seeds 5-8, t +0.24), which is the boot-level check that the other two are dormant here.

### DIFF
--- a/patches/vllm-pr54282-draft-gumbel-salt.patch
+++ b/patches/vllm-pr54282-draft-gumbel-salt.patch
@@ -1,0 +1,405 @@
+Backport of vLLM PR #54282 (commit fe755c889, "Decouple the draft's gumbel noise stream
+from the target's", Aug 29 2026) onto 0.28.0 plus this repo's patch set.
+
+With draft_sample_method=probabilistic, the Gumbel noise that re-samples a rejected draft
+token was drawn from the same Philox offset (seed, position) as the noise that produced
+that draft token, so the verifier's draw was correlated with the draft's and the accepted
+stream was biased toward the drafter. The fix salts the draft's offset (1 << 30) so the
+two streams are independent, and moves the "position before the sampled token"
+convention out of sample_draft into each speculator call site. Greedy draft sampling
+never calls gumbel_sample for drafts and is unaffected.
+
+Upstream's test tests/v1/spec_decode/test_rejection_sampler_utils.py::
+test_gumbel_drafted_rejection_sample_is_unbiased fails on this tree before the patch and
+passes after it (shipped image, RTX 4090). The DFlash2 candidate-selector hunk did not
+apply against this tree (dflash2-lookup-drafting.patch changed its context) and is
+ported by hand: IS_DRAFTING=True and the P-1 position, as at the other call sites.
+Upstream's other test file exercises PR #53017 (draft logits cache column stride) as
+well, which this tree does not carry; that one test is the only failure left and is
+unrelated to this patch.
+
+--- a/v1/worker/gpu/sample/gumbel.py
++++ b/v1/worker/gpu/sample/gumbel.py
+@@ -13,6 +13,12 @@
+ # attribute is `None`, and `tl.constexpr(...)` would crash at import time.
+ _TL_RAND_MIN = tl.constexpr(4.6566127342e-10) if HAS_TRITON else 4.6566127342e-10
+ 
++# Offset salt keeping the draft's Gumbel noise disjoint from the target's.
++# Verification is a probability-ratio test, not a Gumbel coupling, so a proposal
++# and the residual it is resampled from must not share a noise vector.
++# Positions are int64 and never approach 2**30, so the streams cannot collide.
++_DRAFT_NOISE_SALT = tl.constexpr(1 << 30) if HAS_TRITON else (1 << 30)
++
+ 
+ @triton.jit
+ def _temperature_kernel(
+@@ -89,6 +95,7 @@
+     seed,
+     pos,
+     temp,
++    IS_DRAFTING: tl.constexpr,
+     USE_FP64: tl.constexpr,
+     APPLY_TEMPERATURE: tl.constexpr = True,
+ ):
+@@ -108,6 +115,8 @@
+     if USE_FP64:
+         logits = logits.to(tl.float64)
+     if temp != 0.0:
++        if IS_DRAFTING:
++            pos = pos + _DRAFT_NOISE_SALT
+         gumbel_seed = tl.randint(seed, pos)
+         if USE_FP64:
+             u = tl_rand64(gumbel_seed, keys, includes_zero=False)
+@@ -135,6 +144,7 @@
+     logits_cache_stride,
+     logits_cache_col_ptr,
+     vocab_size,
++    IS_DRAFTING: tl.constexpr,
+     APPLY_TEMPERATURE: tl.constexpr,
+     USE_FP64: tl.constexpr,
+     PER_TOKEN_COL: tl.constexpr = False,
+@@ -171,6 +181,7 @@
+         seed,
+         pos,
+         temp,
++        IS_DRAFTING=IS_DRAFTING,
+         USE_FP64=USE_FP64,
+         APPLY_TEMPERATURE=APPLY_TEMPERATURE,
+     )
+@@ -193,6 +204,7 @@
+     temp_ptr,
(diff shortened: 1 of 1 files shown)


mhenrichsen:
Merging, and I want to record why the acceptance question does not hold it.

**What this fixes is a distribution defect, not a performance knob.** Under `draft_sample_method=probabilistic`, which is what this launcher ships since `0e95195`, the noise that re-samples a rejected draft token came from the same Philox offset that produced the draft token. The accepted stream therefore leaned toward the drafter, and upstream's own test puts a number on it: chi2 1584.9 against a threshold of 69.8 at position 0. Speculative decoding's entire contract is that the output distribution is the target's. Without this patch, on this launcher's default, it is not. That is worth a few percent.

**Your 3090 row is probably stronger evidence than you gave it credit for, and it points the same way.** You caveat it:

> that box runs a native tree (cb5679d) on which `verify.sh` reports `dflash2-lookup-drafting.patch` and `dflash2-prewarm.patch` as not applied, before and after this patch alike, so its two arms pair cleanly against each other but not like-for-like against the image tree.

Those two patches, specifically, are the two whose `.orig` file-creation hunks #92 just removed — and that is very likely all that was wrong. `patches/_check_applied.py` builds its file list from every `+++` line in a patch, so the `.orig` backup paths were files verify.sh *required to exist in the installed tree*. Anyone who ran a `find -name '*.orig' -delete` cleanup got exactly those two reported as not applied on a tree that is correctly patched. I reproduced it:

```
dflash2-lookup-drafting: applied(0)      # patched tree, .orig present
dflash2-prewarm: applied(0)
--- find -name '*.orig' -delete ---
dflash2-lookup-drafting: NOT applied(1)  # same tree, verify.sh now lies
dflash2-prewarm: NOT applied(1)
```

If that is what happened on threadchip, the 3090 arms were like-for-like after all, and the -0.121 tok/step paired difference is a real signal on our reference card rather than a tree artifact. Which is fine: **that is the sign you expect from removing a bias toward the drafter.** The drafter was being accepted more often than it deserved; now it is not. A correctness fix that costs acceptance is the correctness fix working, and I would have been more worried by a patch of this shape that cost nothing anywhere.

So: pooled -0.052 tok/step at n=127, t -1.10, unresolved. I am taking it. If it turns out to be a consistent 3% on the 3090 that is a fact about our card and belongs in the docs, not a reason to serve a biased distribution.

The documentation here is exactly right and I want to name the parts, because they are the parts most PRs skip:

- You reported the red honestly, including that the shipped test cannot produce it (`TypeError` before it measures anything) and exactly what you removed from the helper to see the bias on an unpatched tree.
- You accounted for both remaining test failures, traced them to vLLM #53017, and explained why that bug cannot reach this drafter rather than waving at it.
- You explained the missing row (seed 7, prompt 5) and showed it is a property of the prompt across twelve earlier arms.

One follow-up, no rush and not a condition: the hand-ported hunk. The DFlash2 candidate-selector site did not apply because `dflash2-lookup-drafting.patch` moved its context, and you ported it by hand to `IS_DRAFTING=True` and the P-1 position. That is the one line in this patch that upstream did not write and that no upstream test covers on our tree. A sentence in the patch preamble saying which hunk is ours and why would help whoever regenerates this against 0.29.

Issue #87, opened 2026-09-08
say Thanks , A request


jikmiko:
Hello,

First, thank you for your excellent optimization work. I’m getting more than 190 tokens per second at 250 watts on my RTX 3090, which is amazing.

Can this optimization be used with other W4A16 models as well? Is there a README or guide explaining how to do this?

For example, would it work with this model?

https://huggingface.co/JC1DA/Qwen3.8-27B-DavidAU-TURBO-Fable-Cold-Fusion-735-882-Heretic-Uncensored-NM-DAU-INT4-W4A16

Alex-un:
Hey bro, I’m curious how you manage to hit 190 tokens per second at 250 watts. @jikmiko

Dewamade:
> Hello,
> 
> First, thank you for your excellent optimization work. I’m getting more than 190 tokens per second at 250 watts on my RTX 3090, which is amazing.
> 
> Can this optimization be used with other W4A16 models as well? Is there a README or guide explaining how to do this?
> 
> For example, would it work with this model?
> 
> https://huggingface.co/JC1DA/Qwen3.8-27B-DavidAU-TURBO-Fable-Cold-Fusion-735-882-Heretic-Uncensored-NM-DAU-INT4-W4A16

I just check the model you give it using W4A16 auto_round:auto_gptq i think it didn't work for requant the lm_head, the requant script targeted is W4A16 compressed tensor

jikmiko:
@Alex-un  Yeah, that's real. In agentic work, I got 80 to 120 tokens/second with Claude Code with a configured environment, MTP, and long context (MAX_LEN=110000, Prefix_cache=1,vision=0,tools=1,tools_parser=qwen3_coder) with a 3090 using 245W.

With Jan Chat, I got 190 tokens/second, and with Dflash2, I got higher throughput, but it wasn't stable for agentic work.

jikmiko:
@Dewamade Thanks.
I checked this fine-tuned model, and it's really well done. I think it has more intelligence and performs better on one-shot tasks.
This model with llama-cpp and MTP gives me 40 to 50 tokens/second, and I wish I had this optimization for this model.

persano:
Hi guys, just saw this thread and I have some questions regarding the values you are getting!

[...]

Issue #86, opened 2026-09-07
spec-decode-attn: illegal memory access (Xid 31) on long cached-prefix requests — int32 overflow of the block index in _spec_attn_partial


astrapro40k:
**Symptom.** With `VLLM_SPEC_DECODE_ATTN=1` and a DFlash2 draft (K=7), a request of ~200k tokens whose first ~158k came from the prefix cache kills the engine on the first verify step after prefill:

```
torch.AcceleratorError: CUDA error: an illegal memory access was encountered
NVRM: Xid (PCI:0000:01:00): 31, ... MMU Fault: ENGINE GRAPHICS GPC0 ... FAULT_PDE ACCESS_TYPE_VIRT_READ
```

Reproduced on vLLM 0.27.1 (this repo's earlier patch set) and on 0.28.0 with the current `patches/*.patch`, target Qwen3.8-27B (W8A16 attn / W4A16 MLP, hybrid GDN), draft `Qwen3.8-27B-DFlash2-tuned2-W4A16`, one CMP 170HX 64 GB (sm80), `--max-model-len 262144 --gpu-memory-utilization 0.9` (KV pool ≈ 480k tokens). Sequence that triggers it every time: 128k cold prompt → 160k prompt sharing the prefix (80% hit) → 200k prompt sharing the prefix (79% hit). `VLLM_SPEC_DECODE_ATTN=0` makes the same sequence pass, so it is the split-KV kernel.

**Cause.** In `_spec_attn_partial` the block id is loaded from the int32 block table and multiplied by `stride_kb` / `stride_vb` in int32:

```python
blk = tl.load(bt_ptr + req * stride_bt + pos // BLOCK_SIZE, mask=k_ok, other=0)
k_ptrs = k_ptr + blk[:, None] * stride_kb + ...
```

With a large pool the product overflows for high block ids, which is exactly what a request that reuses a long cached prefix gets for its *new* blocks (the cold 128k request sits low in the pool). On a 24 GB card the pool is small enough that this never shows.

**Fix** (one line after the load):

```diff
         blk = tl.load(bt_ptr + req * stride_bt + pos // BLOCK_SIZE, mask=k_ok, other=0)
+        # int64: blk * stride_kb overflows int32 for high block ids in a large KV pool
+        blk = blk.to(tl.int64)
         slot = pos % BLOCK_SIZE
```

[...]

PR #90, opened 2026-09-09
Split-KV verify attention on the per-tensor fp8 KV cache (TRITON_ATTN, sm89+)

cpuchip:
The split-KV verify kernel (`patches/spec-decode-attn.patch`) reads vLLM's per-tensor fp8 KV cache, so `--kv-cache-dtype fp8` can run with `TRITON_ATTN` on both the target and the drafter and keep FULL CUDA graphs. sm89 and up: Triton has no fp8e4nv conversion on sm86, so on a 3090 the route does not exist and nothing here changes what that card runs.

Rebased onto main after #91 merged, so the diff is this patch's own files: `patches/triton-spec-attn-fp8-kv.patch` (the kernel's fp8 path in `spec_decode_attn.py`, the `TRITON_ATTN` and `FLASH_ATTN` backend hooks, and two `envs.py` hunks registering `VLLM_SPEC_ATTN_DEBUG`, the kernel's debug print switch, which was read with a bare `os.environ.get` and never registered; all three sites now read it through `envs`, since a raw read of "0" is truthy and would have turned the debug path on), `patches/series` (tail line), `bench/test_spec_decode_fp8.py` (correctness against a dequantized reference and timing against the bf16 kernel and FA2; it now skips by design below sm89 and on a kernel without the fp8 path instead of dying in the compiler), `docs/long-context.md` (the route, its verbatim launch line, the native rows) and `docs/gotchas.md` item 55 (the route's two limits: sm89 and up, and `INT8_ACT=int8` does not stack on it).

Answering the review's three asks: the capability guard is in; the gotcha is item 55; merge order was #91 first, and it is in, so this branch no longer carries #91's commits.

**Measured on a WSL2 4090 at 120k**, DFlash2 k=7: fp8 decode 108.7 / 88.6 / 81.0 tok/s at 24k / 49k / 88k tokens with this patch against 71.4 / 50.9 / 35.4 with vLLM's unified attention on the same route, and 84.4 / 81.7 / 83.5 for FlashInfer with PIECEWISE (issue #87). **Native 4090** (Ubuntu 24.04, driver 580.159, this image), prose prompts of about 24k / 50k / 89k tokens: 100.6 / 96.0 / 87.4 tok/s against 68.3 / 49.8 / 34.6 with unified attention (median of 3, greedy, 256 out, fresh prefill, `MAX_LEN=120000 PREFIX_CACHE=1 MAX_SEQS=4`); prefill unchanged by the patch on either box (2.4k / 2.0k / 1.6k tok/s native). Correctness: the 13-shape oracle in `bench/test_spec_decode_attn.py` is unchanged by this patch (the fp8 constexprs specialise away on bf16, and the maintainer measured int64+fp8 within 0.5% of int64-only at every shape on the 3090); the fp8 path's worst case against a dequantized reference is 0.0084 on the 4090.

**On the 3090:** `bench/test_spec_decode_fp8.py` prints its skip line and exits 0, which verifies the guard and not the kernel; nobody on an sm86 card can exercise the fp8 path. From the maintainer's runs: `patch -p1 --dry-run` of the patch against the live tree applies with no fuzz and no offsets; the bf16 half of `bench/test_spec_decode_bigpool.py` shows the #86 fault before the cast and `BIGPOOL OK` after; the fp8 half of that test skips by design on sm86 and, on main without this patch, on any card, because the kernel has no `k_descale` argument; with this patch merged it exercises the fp8 path on sm89 and up. End-to-end A/B at 25k (`bench/labd_bench.py`, two paired reps per arm): no detectable change on that card, as expected for a route it cannot take.

**Knob sense, verified on the native 3090 against this head:** the same profile and request with only the knob differing, `VLLM_SPEC_ATTN_DEBUG=1` prints 19 "syv spec-attn" lines and `=0` prints none (the two `spec_decode_attn.py` sites fire on the FLASH_ATTN path the 3090 runs; the `triton_attn.py` site is gated on `TRITON_ATTN` and a multi-query verify); the registry reads "0" as False and "1" as True. Before this fix a raw read of "0" was truthy and turned the debug path on.

**Series:** in `patches/series` order the whole series applies to a pristine v0.28.0 checkout with this patch landing on exact context in its tail position (`check_vllm_series.sh`: 33 applied, 0 rejected).

**What is still unproven:** the 108.7 / 88.6 / 81.0 and the 0.0084 worst case rest on the 4090 builds and nobody on a 3090 can move them either way (every fp8 path compile-eliminates on sm86); not tested under the CUDA-graph-captured path with prefix caching on beyond the labd run; `INT8_ACT=int8` on this route is slow with or without the patch and the cause (a Marlin variant choice) is tracked separately; registering the debug knob puts it in vLLM's registry the way the precedent patch does, and whether that changes the compile-cache hash on this vLLM is argued from the precedent, not measured.


cpuchip:
The `INT8_ACT=int8` caveat in the description is resolved, and it is memory, not the kernel.

Same box, same route (fp8 cache, Triton on both sides, this kernel), same int8-activation setting, one 4k request each, decode tok/s and Windows' own shared-GPU-memory counter for the card (host RAM standing in for VRAM):

| max length | decode | fresh prefill | shared GPU memory during decode |
|---|---|---|---|
| 120,000 | 44.0 | 2044 | 296 MB (0 before the boot) |
| 60,000 | 123.9 | 3599 | 90 MB |

A torch profile at 120k shows the same Marlin kernels as a bf16 boot, same templates and call counts, running 2x to 100x slower (int8 decode GEMM 103-143 us against 57.5; the 16-bit lm_head 2.4-3.7 ms per call against 0.5, 132 ms per prefill call against 1.4). Under WSL2 an oversubscribed card does not fail the allocation the way Linux does; WDDM pages it to host memory, and a kernel that touches paged weights is exactly that signature. The untouched image tree on the same route shows the same numbers, so this predates the patch; the patch only removed the attention time that had been hiding it.

The rule for this route on a 24 GB card: leave headroom when stacking `INT8_ACT=int8` on the fp8 pool at 120k (a shorter max length, a smaller KV pin, or a lower utilization; 60k was enough here), or run the fp8 route without int8 activations at 120k, which is the 108.7 / 88.6 / 81.0 row in the description. A native Linux box would surface the same overcommit as an OOM at boot rather than a slow server, which is the better failure; that row will come with the native 4090 numbers.

mhenrichsen:
Held on #91, not on anything here. This branch carries `a9d9442` and that commit does not contain the int64 cast (details in #91) — it adds only `bench/test_spec_decode_bigpool.py`. Since you say this kernel applies on top of `spec-decode-attn.patch` *with* the cast, the stated precondition is not actually met by the branch. Fix #91, rebase, and I will take this.

What I could check without an sm89 card, all green:

- `patches/check_vllm_series.sh` against a pristine v0.28.0: **31 patches, 0 offsets, both passes, `patch integrity: OK`** with `triton-spec-attn-fp8-kv.patch` in the glob. The name sorts after `spec-decode-int8-kv` and `spec-decode-scratch-within-budget`, so the order you rely on is the order the loop produces.
- The gate is genuinely conditional: `spec_fp8` requires `self._kv_quant_mode == KVQuantMode.FP8_PER_TENSOR` **and** `key_cache.dtype == self.fp8_dtype`. On the shipped route (int8 or bf16 KV, no `--kv-cache-dtype fp8`) nothing in the patch can be reached, which is what makes a patch for hardware we do not own safe to carry.

The documentation here is the standard I want on this queue — the control rows on the 3090 in particular, where the claim is "this is a no-op" and the evidence is byte-identical correctness lines and bit-identical C1 generation rather than an assertion. Three things before merge, none of them large:

1. **The native sm89 row you promised.** Every number above is WSL2, and this repo has been bitten more than once by WSL2-only effects. I am not asking you to rent a 4090 twice; if the rented row is not coming, say so and I will merge it labelled WSL2-only, with that written into `docs/long-context.md` next to the table.
2. **`docs/long-context.md` gains 10 lines but the route does not gain a launcher tier**, so the only record of the exact invocation is a PR body. Put the full `EXTRA_ARGS` line in the doc, verbatim, including the `"attention_backend":"TRITON_ATTN"` inside the speculative config. That inner setting is the part someone will drop, and dropping it silently costs them FULL graphs.
3. **The `INT8_ACT` interaction needs to be in the doc, not just here.** "These two settings should not be stacked until the memory-pressure question is understood" is a real operational warning and it is currently one paragraph in a PR body. One sentence in `docs/long-context.md` and, if you want it durable, a gotcha.

On the profiling paragraph: identical kernels, identical call counts, 2-100x slower, at 95% card occupancy is a convincing case that it is not kernel selection. I agree it belongs in its own issue rather than blocking this.

cpuchip:
Rebased on the corrected #91 (the int64 cast is now in `spec-decode-attn.patch`, so the precondition this branch relied on is actually met). Taking your three, none large:

1. **The native sm89 row.** I do not have a native 4090 in hand, and rather than hold this on a rental I would rather you merge it labelled WSL2-only, with that written into `docs/long-context.md` next to the table. Every number in the PR is WSL2 and I have said so; if a native row comes later I will add it.

2. **The verbatim `EXTRA_ARGS` line in the doc.** Added to `docs/long-context.md`, including `"attention_backend":"TRITON_ATTN"` inside the speculative config, called out as the setting that silently costs FULL graphs if dropped.

3. **The INT8_ACT interaction in the doc.** Added: the one sentence that these two settings should not be stacked until the memory-pressure question is understood, plus a gotcha.

On the profiling paragraph: agreed it belongs in its own issue rather than blocking this, and I will open one.

cpuchip:
The native sm89 row is in, so the WSL2-only label can come off. Rented bare-metal RTX 4090 (Ubuntu 24.04, kernel 6.8, driver 580.159), this image at `latest` with the #91 cast applied, the launch line from `docs/long-context.md` verbatim (`SPEC=dflash2 CTX=fast`, TRITON_ATTN on both sides, fp8 KV, k=7), `MAX_LEN=120000 PREFIX_CACHE=1 MAX_SEQS=4`, fresh prefill each request, 256 out, greedy, median of 3, prose prompts from Project Gutenberg:

| prompt tokens | unified attention verify (image as shipped) | split-KV verify (this patch) | ratio |
|---|---|---|---|
| ~24k | 68.3 tok/s decode, 2,402 prefill | 100.6 decode, 2,394 prefill | 1.47x |
| ~50k | 49.8, 2,004 | 96.0, 1,982 | 1.93x |
| ~89k | 34.6, 1,580 | 87.4, 1,582 | 2.53x |

Same shape as the WSL2 row (71.4 / 50.9 / 35.4 against 108.7 / 88.6 / 81.0 at 24k / 49k / 88k): the unpatched arm is within 5% of WSL2 at every depth, the patched arm within 8%, and prefill is untouched by the patch on both boxes. The 138,214-token fp8 pool at `MAX_LEN=120000` also matches. So nothing in the earlier numbers was a WSL2 effect, and I have updated the doc paragraph to say native and WSL2 rather than WSL2-only. The runner script and the per-rep rows are available if you want them.

mhenrichsen:
Holding this one short of merge, on two bookkeeping items and one small bench fix — not on the engineering, which checks out on the box.

What I verified on the 3090 (production stopped, power limit untouched at 250 W, tree `~/q28` = the 0.28.0 production install, PR fetched as `pull/90/head`, head 1086ed2):

**The int64 cast is a real bug fix on our card, not just on a 64 GB one.** Stock tree, your new bench:
```
python bench/test_spec_decode_bigpool.py
RuntimeError: Triton Error [CUDA]: an illegal memory access was encountered   (exit=1)
```
After applying the branch: `bf16: pool 2600 blocks, request on blocks 2577..2599, L=20000: max|ours-ref|=0.0001 OK`. So #86 is live on main here too, and a9bbbbf genuinely fixes what a9d9442 only titled — thanks for turning that around cleanly.

**Patch application:** `patch -p1 --dry-run` of `patches/triton-spec-attn-fp8-kv.patch` against the live tree — three files, no fuzz, no offsets. CI git-apply green.

**Offline correctness unchanged:** 13-shape oracle PRE vs POST, correctness lines byte-identical (13 OK lines both sides).

**End-to-end A/B**, `bench/labd_bench.py` at 25k ctx, two paired reps per arm, arm A = branch, arm B = stock. Warm-prefix rep matched on all six tasks and `TOTAL out=2685 steps=805 tok/step=3.35 decode=105.1 vs 105.0 tok/s`. Cold-prefix rep matched on five of six; `code` diverged (71 vs 82 steps) but `code` also flips within arm A between its own reps, i.e. the known near-tie behaviour of spec decode, not your patch. Within-arm spread (109.1 vs 105.1) dwarfs between-arm, so read this as "no detectable change", not a win.

**Cost of the cast, isolated** (`bench/test_spec_decode_attn.py`, 3 reps/arm, idle GPU, medians): int64+fp8 is within 0.5% of int64-only at every shape, so the fp8 constexprs really do specialize away on bf16 — your dead-branch claim holds. The cast itself costs +1.0 to +2.6% at q=8 (consistent 3/3) and *gains* 1.0–3.5% at q=16 and q=32. Production runs DFLASH_TOKENS=15 → q=16, the faster side. I did not chase why; likely Triton register allocation. Either way it is ~10% of step time, so the q=8 figure is ~0.25% of a step.

What I cannot confirm, and am not asking you to prove here: the 108.7/88.6/81.0 vs 71.4/50.9/35.4 headline and the fp8 correctness worst-case 0.0084. Every fp8 path compile-eliminated on sm86 — the Triton compiler refuses fp8e4nv outright (gotchas.md gotcha 40). Those numbers rest on your build and nobody on this card can move them either way. That is fine for a gated sm89+ path; I am recording it, not disputing it.

Three things to unblock the merge:

1. `bench/test_spec_decode_fp8.py` has no device-capability guard. On the 3090 it does not skip, it dies with a raw `CompilationError: type fp8e4nv not supported in this architecture`. Its sibling `test_spec_decode_bigpool.py` handles exactly this and prints `fp8: SKIP(by design)`. Same guard in the fp8 bench, please — otherwise the bench dir is red on every 3090 for a file that is working as designed.

2. Review ask 3 is half done. The INT8_ACT-on-fp8 stacking warning is in `docs/long-context.md`, but the gotcha you promised never landed — `docs/gotchas.md` is not in this branch's diff. Add it.

3. Merge order: this branch carries both of #91's commits, so merging #90 merges #91. Tell me which you want — #91 merged first (then this diff shrinks to its own three files), or #91 closed as absorbed here. I will not merge #90 and then #91 separately.

Asks 1 and 2 from the review thread are discharged: the native bare-metal 4090 row is in the thread, the doc says native and WSL2, and the verbatim EXTRA_ARGS line with the inner `"attention_backend":"TRITON_ATTN"` is in `docs/long-context.md`. `docs/MR-DRAFT.md`'s owed checks are about the VLLM_INT4_MQ_3D default flip and are out of scope here — this PR flips no defaults.

Push the guard and the gotcha, answer the merge-order question, and I will merge.

cpuchip:
Pushed, the three asks and two more found on the way:

1. `bench/test_spec_decode_fp8.py` has the same two guards as the bigpool sibling; on a 3090 it prints one `fp8: SKIP(by design)` line and exits 0 (verified on a native 3090 against this head).
2. Gotcha 55: the route is sm89 and up, and `INT8_ACT=int8` does not stack on it.
3. Merge order: #91 first, and it is in, so this branch is rebased onto 8d09ec5 and carries only its own files: the patch, its `patches/series` line, the bench, `docs/long-context.md`, gotcha 55.

Found while doing that: the kernel's debug switch `VLLM_SPEC_ATTN_DEBUG` was read with a bare `os.environ.get` and never registered, so it sat outside `envs.py` and the compile cache key, and a raw read of "0" is truthy. The patch now registers it in `envs.py` (two hunks after `speed-knobs-envs.patch`) and reads it through `envs` at all three sites. Verified on the native 3090 with the same profile and request and only the knob differing: `=1` prints 19 debug lines, `=0` prints none; the registry reads "0" as False. Series integrity at this head against a pristine v0.28.0: 33 applied, 0 rejected, the fp8 patch exact at the tail (16 hunks, 4 files).

Still unproven: the fp8 kernel itself on sm86, where the bench correctly declines to run, so the 3090 rows verify the guard and not the kernel; the 4090 numbers stand on the 4090 builds as before.

mhenrichsen:
Merging. The three round-two asks are all in, the 3090 says this is a no-op on sm86 exactly as you claimed, and I absorbed the remaining chores rather than bouncing them back.

**What I verified on our card** (production stopped, GPU idle at 18 MiB, power limit untouched at 250 W):

- `patch -p1 --dry-run` against the deploy tree raised to main-equivalent: four files, zero offsets, zero fuzz.
- `bench/test_spec_decode_fp8.py` pre- and post-patch: `fp8: SKIP(by design): no fp8 conversion on this card (sm86; needs sm89+)`, exit 0 both times. The guard works; the kernel is untested here and cannot be tested here.
- `bench/test_spec_decode_attn.py`, both arms in one boot on the same warm Triton cache: 13/13 OK, 0 FAIL, and the 13 `max|ours-ref|` lines are byte-identical pre vs post (`diff` -> no output). Kernel timings us/layer pre -> post: 1500/q8 29.2 -> 29.4, 1500/q16 54.3 -> 54.2, 4000/q16 110.7 -> 110.5, 25000/q16 552.9 -> 556.3, 60000/q32 2662.2 -> 2663.8. The one outlier, 60000/q16 1323.6 -> 1402.0, is inside that shape's own run-to-run spread (an earlier run of the same pre-patch binary printed 1366.8).
- `bench/test_spec_decode_bigpool.py` with the fp8 patch on top: `bf16: ... max|ours-ref|=0.0001 OK`, fp8 half `SKIP(by design)`, `BIGPOOL OK`.
- Booted the server on the patched tree, greedy completion correct, and `journalctl --user -u qwen-single | grep -E "spec-attn|fp8|error|Traceback"` over the whole boot printed nothing: the `FP8_PER_TENSOR and key_cache.dtype == self.fp8_dtype` gate is never reached on the int8/bf16 route we ship. Then reverse-applied and md5-compared all four files against a pre-flight backup: identical, service back up, healthy.
- Your truthiness fix is real and I measured it: `VLLM_SPEC_ATTN_DEBUG=1` -> 6 "syv spec-attn" lines, `=0` -> 0, `""` -> 0, unset -> 0. The previous revision (`git show 037140b:patches/triton-spec-attn-fp8-kv.patch`) read it bare at three sites, so `=0` turned the debug path on. That is the part of this PR that touches code the 3090 actually executes, and it is a fix, not a risk.

**What I changed on your branch** (maintainer edits, four commits):

1. Dropped the dead `+import os` from the `triton_attn.py` hunk. With all three reads going through `envs`, nothing in that file uses `os` (`grep -c "\bos\." triton_attn.py` -> 0 on the patched tree). Removing the added line left the hunk pure context, which GNU patch rejects, so the hunk went with it and the two after it got their new-file start renumbered.
2. Rebased onto main. `patches/series` and `docs/gotchas.md` both conflicted — sse-keep-alive and #93's int4 item landed meanwhile — so the fp8 gotcha renumbers from 55 to **57** and the series line stays at the tail.
3. Re-cut the two `envs.py` hunks onto `int4-mq3d-envs.patch`: #93 put `VLLM_INT4_MQ_3D` / `VLLM_INT4_MQ_3D_DEBUG` in both of the regions your hunks anchored on.
4. Regenerated those hunks with `diff -u` after hand-cutting them failed. Worth knowing, because it cost me an hour: my hand-cut hunks matched the post-series file byte for byte and `git apply --check` accepted them, but GNU patch failed hunk 1 and took hunk 2 at fuzz 1. `patches/series` is defined against GNU patch, so GNU patch is the referee — always cut series hunks with `diff -u` against the tree the series produces.

Series check green after all of that, on Linux with GNU patch 2.7.6 against a pristine v0.28.0 checkout: `bash patches/check_vllm_series.sh <pristine v0.28.0>` -> **38 patches applied, patch integrity: OK**, with this patch landing at the tail on exact context. CI git-apply is green on the head I merged.

**Still unproven, and stated as such in the merge:** the 108.7 / 88.6 / 81.0 and the 0.0084 worst case are your 4090 numbers and nobody on sm86 can move them in either direction — every fp8 branch compile-eliminates here. We are merging a route we cannot execute, on the strength of your measurements plus a proof that it is inert on our card. That is the right trade for a patch that is gated this tightly, but the docs rows should keep saying whose hardware produced them. Thanks for the native bare-metal 4090 row — that is what got the WSL2-only label off this.

One thing your fix exposed, filed as #114 rather than held against you: `patches/spec-decode-attn.patch` still reads `VLLM_SPEC_DECODE_ATTN_QMAX` (line 109) and `VLLM_SPEC_ATTN_BLOCK_M` (line 298) with bare `os.environ.get`, so they sit outside the registry and the torch.compile cache key — same class as the knob you just fixed, and worse, since `BLOCK_M` changes the kernel launch shape.

PR #91, opened 2026-09-10
spec-decode-attn: cast the block id to int64 before the stride multiply (fixes #86)

### LINKED ISSUE #86: spec-decode-attn: illegal memory access (Xid 31) on long cached-prefix requests — int32 overflow of the block index in _spec_attn_partial

cpuchip:
Fixes #86. The block id loaded from the int32 block table in `_spec_attn_partial` is multiplied by the cache strides in int32, which overflows for high block ids in a large KV pool (astrapro40k's report: a 480k-token pool on a 64 GB card, illegal memory access on the first verify after a long cached prefix). The fix is the one line proposed there, `blk = blk.to(tl.int64)` after the load, placed in `spec-decode-attn.patch` where the kernel lives; the int8 per-token-head scale pointers index with the same block id, so the one cast covers them.

`bench/test_spec_decode_bigpool.py` reproduces it on a 24 GB card: a 2,600-block pool with one request placed at the top. Before the cast, `Triton Error [CUDA]: an illegal memory access was encountered`, every time, on a 4090 (WSL2) and a native 3090. After it, max error 0.0002 against an fp32 reference on the 4090 (0.0001 on the 3090) for bf16, and 0.0002 for fp8 where that path exists; the existing `bench/test_spec_decode_attn.py` byte-identical on both boxes; per-layer timing within single-run noise. The test's fp8 half has two designed skips and names which one fired, both verified on real hardware: a kernel without the fp8 path prints `SKIP(by design): this kernel has no fp8 path (no k_descale argument)` (both cards), and with the fp8 path present a 3090 prints `SKIP(by design): no fp8 conversion on this card (CompilationError)`; any other exception there fails the run. The later patches that touch this file (`spec-decode-int8-kv`, `spec-decode-scratch-within-budget`) apply unchanged; the cast sits outside their hunk context.

#90 will rebase on this.

### DIFF
--- a/bench/test_spec_decode_bigpool.py
+++ b/bench/test_spec_decode_bigpool.py
@@ -1,0 +1,69 @@
+"""Issue #86: int32 overflow of blk * stride in _spec_attn_partial on a large KV pool.
+A request whose blocks sit high in a pool larger than 2^31 elements / stride_kb reads the wrong memory
+(bf16: stride_kb = 880*4*256 = 901,120 elements, so block ids above 2,383 overflow). Places one request's
+blocks at the TOP of a 2,600-block pool and checks the kernel against the reference; bf16 and fp8 paths.
+Run inside the vLLM venv after the patches are applied (any card): venv/bin/python bench/test_spec_decode_bigpool.py
+The fp8 half has two designed skips (a kernel without the fp8 path; a card without fp8 conversion, sm86) and
+reports which one fired; anything else there fails the run. Exit code 0 = all OK; a CUDA illegal access kills
+the process (that is the 'before' result)."""
+import sys, torch
+from vllm.v1.attention.ops.spec_decode_attn import SpecDecodeAttention
+torch.manual_seed(0); dev = "cuda"
+Hq, Hkv, D, BS = 24, 4, 256, 880
+scale = D ** -0.5
+FP8 = torch.float8_e4m3fn; FMAX = torch.finfo(FP8).max
+NB = 2600  # blocks in the pool: NB*BS*Hkv*D = 2.34e9 elements > 2^31
+att = SpecDecodeAttention(max_num_reqs=8, num_heads=Hq, head_dim=D, device=dev, qmax=64)
+def run(dtype_name, L=20000, q_len=8):
+    nb_req = (L + BS - 1) // BS
+    if dtype_name == "bf16":
+        kc = torch.empty(NB, BS, Hkv, D, device=dev, dtype=torch.bfloat16).normal_()
+        vc = torch.empty(NB, BS, Hkv, D, device=dev, dtype=torch.bfloat16).normal_()
+        k8 = v8 = None; kd = vd = None
+    else:
+        kc = torch.empty(NB, BS, Hkv, D, device=dev, dtype=torch.bfloat16).normal_()
+        s = (kc.abs().amax() / FMAX).clamp(min=1e-6); k8 = (kc / s).clamp(-FMAX, FMAX).to(FP8); kd = s.reshape(1).float(); del kc
+        vc = torch.empty(NB, BS, Hkv, D, device=dev, dtype=torch.bfloat16).normal_()
+        t = (vc.abs().amax() / FMAX).clamp(min=1e-6); v8 = (vc / t).clamp(-FMAX, FMAX).to(FP8); vd = t.reshape(1).float(); del vc
+        torch.cuda.empty_cache()
+        kc = k8; vc = v8   # the reference dequantizes only the request's own blocks below
+    ids = torch.arange(NB - nb_req, NB, device=dev, dtype=torch.int32)   # the highest block ids
+    bt = ids.reshape(1, nb_req)
+    q = torch.randn(q_len, Hq, D, device=dev, dtype=torch.bfloat16)
+    cu = torch.tensor([0, q_len], dtype=torch.int32, device=dev); su = torch.tensor([L], dtype=torch.int32, device=dev)
+    out = torch.empty_like(q)
+    if dtype_name == "bf16":
+        att.run(q, kc, vc, out, cu, su, bt, scale, 1, q_len)
+    else:
+        att.run(q, k8, v8, out, cu, su, bt, scale, 1, q_len, k_descale=kd, v_descale=vd)
+    torch.cuda.synchronize()
+    k = kc[ids.long()].reshape(-1, Hkv, D)[:L].float(); v = vc[ids.long()].reshape(-1, Hkv, D)[:L].float()
+    if kd is not None:
+        k = k * kd; v = v * vd   # dequantized reference values for the request's blocks
+    G = Hq // Hkv; k = k.repeat_interleave(G, 1); v = v.repeat_interleave(G, 1)
+    s_ = torch.einsum("qhd,khd->hqk", q.float(), k) * scale
+    qpos = torch.arange(L - q_len, L, device=dev)[:, None]; kpos = torch.arange(L, device=dev)[None, :]
+    s_ = s_.masked_fill((kpos > qpos)[None], float("-inf")); p = torch.softmax(s_, -1)
+    r = torch.einsum("hqk,khd->qhd", p, v)
+    e = (out.float() - r).abs().max().item()
+    print(f"  {dtype_name}: pool {NB} blocks, request on blocks {NB-nb_req}..{NB-1}, L={L}: max|ours-ref|={e:.4f} {'OK' if e < 0.05 else 'FAIL'}", flush=True)
+    return e < 0.05
+results = [run("bf16")]
+torch.cuda.empty_cache()   # the bf16 pool and its reference are gone with run()'s frame; give the fp8 half a clean card
+# The fp8 half has exactly two designed skips; anything else is a failure of the run, not a skip.
+try:
+    results.append(run("fp8"))
+except TypeError as e:
+    if "k_descale" in str(e):
+        print("  fp8: SKIP(by design): this kernel has no fp8 path (no k_descale argument)", flush=True)
+    else:
+        raise
+except Exception as e:
+    msg = str(e)
+    if "fp8e4nv" in msg and "not supported in this architecture" in msg:
+        print(f"  fp8: SKIP(by design): no fp8 conversion on this card ({type(e).__name__})", flush=True)
+    else:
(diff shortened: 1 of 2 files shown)


persano:
Tested on WSL2 / RTX 3090 (sm86), local image built from the repo Dockerfile with this PR cherry-picked onto `main`; `verify.sh` all PASS. On the int8 route (`VLLM_SPEC_DECODE_ATTN=1`, `--kv-cache-dtype int8_per_token_head`, pool 145,030 tokens) an 80K request with markers at 5K and 60K depth returned both exact, no CUDA fault. Before this patch that route was on our known-issue list because of #86; it isn't now.

mhenrichsen:
Blocking, and I think it is a slip rather than a disagreement: **the cast is not in this branch.** The commit is titled for it, the body describes it, but `a9d9442` adds one file and nothing else.

```
$ git diff --stat origin/main...pr91
 bench/test_spec_decode_bigpool.py | 69 +++++++++++++++++++++++++++++++++++++++
 1 file changed, 69 insertions(+)

$ grep -n int64 patches/spec-decode-attn.patch
(no output)
```

The kernel on main is still

```python
blk = tl.load(bt_ptr + req * stride_bt + pos // BLOCK_SIZE, mask=k_ok, other=0)
slot = pos % BLOCK_SIZE
k_ptrs = k_ptr + blk[:, None] * stride_kb + ...
```

so #86 is live. Push the one line into `patches/spec-decode-attn.patch` and this goes in — the test is the hard part and it is already written and already run on two cards.

Two notes while you are in there:

- The test is the "before" case by construction: on a tree without the cast it dies with an illegal access rather than failing an assert. Worth one line in its docstring saying the expected pre-fix outcome is a killed process, not a red assertion, so nobody reads a crash as a broken test. Your header does say this; I mean say it where the runner sees it, in the failure path.
- #90 carries this same commit and the same test file, so it inherits the gap. Land this first, then rebase #90 on the corrected commit rather than the current one.

The rest of the reasoning is right and I checked the arithmetic: bf16 `stride_kb` is 880 * 4 * 256 = 901,120 elements, so block ids above 2,383 overflow int32, and a 2,600-block pool with the request placed at the top clears that by 217 blocks. The int8 per-token-head scale pointers index with the same `blk`, so the one cast does cover them.

cpuchip:
You are right, and thank you for catching it. The commit was titled and described for the cast but only carried the test file, so #86 stayed live. That is a delivery gap, not a disagreement.

The cast is now in `patches/spec-decode-attn.patch`:

```
blk = tl.load(bt_ptr + req * stride_bt + pos // BLOCK_SIZE, mask=k_ok, other=0)
blk = blk.to(tl.int64)
```

Falsified on a 4090 before pushing: without the cast `bench/test_spec_decode_bigpool.py` dies with the illegal access; with it, `BIGPOOL OK`, max error 0.0002. The bench also gained a line in its pre-crash path so a killed process reads as the pre-fix outcome rather than a broken test, as you asked.

The cast covers the int8 per-token-head scale pointers too, since they index with the same `blk`, and your arithmetic is the one I used: bf16 `stride_kb` = 880 * 4 * 256 = 901,120 elements, block ids above ~2,383 overflow int32, and the 2,600-block pool with the request at the top clears that by 217.

#90 is rebased on the corrected commit.

persano:
Verified on WSL2 / RTX 3090 (sm86). Local image `qwen-syv:local-2026-09-11` = `origin/main@1834917` + this PR's two commits (`a9d9442` test, `a9bbbbf` cast) + #85 + #55; `verify.sh` OK, 0 failures.

`bench/test_spec_decode_bigpool.py` (2,600-block pool, request on blocks 2577..2599, L=20000):

bf16: max|ours-ref|=0.0001 OK
fp8: SKIP(by design): this kernel has no fp8 path (no k_descale argument)
BIGPOOL OK


Exit 0, no illegal memory access, the pre fix failure mode. Run in a temp container; a server side pool that size OOMs on 24 GB.

End to end on the live int8 route (`VLLM_SPEC_DECODE_ATTN=1 --attention-backend TRITON_ATTN --kv-cache-dtype int8_per_token_head`, pool 145,030 tokens): an 80K prompt with markers at ~5K and ~60K depth returns both exact, 209 s cold; warm recheck of the same prompt 4.4 s with 79,152/80,082 tokens cached, no fault. So the cast holds at high block ids on a real cached prefix, not just in the bench.

mhenrichsen:
Merging.

The branch now contains the change the title advertises — a9d9442 was test-only, a9bbbbf adds the cast — and I reproduced #86 and its fix on a 3090 with production stopped.

Repro, on the pre-fix installed kernel (`vllm/v1/attention/ops/spec_decode_attn.py`, no `int64` anywhere):

    $ python test_spec_decode_bigpool.py
    RuntimeError: Triton Error [CUDA]: an illegal memory access was encountered   EXIT=1

With your cast inserted into that same file (`__pycache__` and the Triton cache wiped, fresh process):

    bf16: pool 2600 blocks, request on blocks 2577..2599, L=20000: max|ours-ref|=0.0001 OK
    BIGPOOL OK   EXIT=0

`bench/test_spec_decode_attn.py` reports okcount=13 on both arms, with max|ours-ref| matching max|FA-ref| to the printed digit on every case. Timing A/B from the same bench, arms alternated within one session, 3 reps each, both Triton cache entries pre-warmed and discarded: at the production geometry (kv=60000, q_len=8) the cast arm is +0.98%; across the 12 geometries the deltas run +3.15% to -3.30% with mixed signs, i.e. noise, not a systematic cost. That mattered to check because `single-user/start_qwen.sh` sets `VLLM_SPEC_DECODE_ATTN=1`, so this is the spec-verify attention on the default path, not an opt-in.

Series integrity I ran locally against a pristine v0.28.0 checkout rather than trusting CI: `bash patches/check_vllm_series.sh` → 30 patches applied, 0 with an offset, git-apply pass clean.

What is still unproven, for the record:

- The fp8 half of your reproducer SKIPs on this install by design (no `k_descale` argument; we carry the int8-KV patch instead). That the one cast also covers the quantized scale strides is verified by reading — the installed kernel's `ks_ptr + blk * stride_ksb` load uses the same `blk` — and not by a run. Nobody has exercised the int8 scale-stride path at high block ids on any card.
- The bug is unreachable in this repo's own 24 GB config: the pool is ~66 blocks (57,669 tokens / 880) against an overflow threshold of 2^31 / (880*4*256) ≈ 2,383 blocks. So this run proves the kernel is buggy and that your cast fixes it; it does not show the 3090 deployment was ever affected. The fix is for large-card users, which is how I'd frame it in any release note.
- Not tested under the CUDA-graph-captured path or with prefix caching on — standalone kernel only.

Credit where it's due: the synthetic reproducer is the reason this was a ten-minute verification instead of a day of guessing. Forcing the blocks to the top of an oversized pool is exactly the right way to make a 64 GB bug fire on a 24 GB card, and the arithmetic in the issue thread was right.

Two non-blocking follow-ups after this lands:

1. `docs/gotchas.md` item 12 covers spec-decode-attn.patch and its query-cap trap but says nothing about the large-pool limit. One sentence — the cast, and that a pool over ~2,383 blocks was silently reading the wrong memory before it — keeps that file honest.
2. #90 needs to sit on a9bbbbf, not a9d9442, so it doesn't inherit the gap. You said that's done; I'll re-check once this is in.

persano:
On the two paths you list as untested: the 80K probe I ran was on the live server with the int8 KV route (`VLLM_SPEC_DECODE_ATTN=1 --attention-backend TRITON_ATTN --kv-cache-dtype int8_per_token_head`), prefix caching on and graphs captured. Markers at ~5K and ~60K depth both exact, warm recheck 4.4 s with 79,152/80,082 tokens cached, no fault. Block ids stayed well under the threshold on a 145,030 token pool, so it doesn't exercise the overflow, but it does cover the int8 scale stride path and the captured graph path that the standalone bench leaves out.

You're also right that the 24 GB config can't reach the bug. I'll correct my README, which had the int8 profile flagged as affected.

PR #92, opened 2026-09-10
Remove .orig backup files from two patch files

jnohlgard:
## Summary

`dflash2-lookup-drafting.patch` and `dflash2-prewarm.patch` contain file-creation
hunks for `.orig` backup files: `qwen3_dflash.py.orig` (899 lines),
`model_runner.py.orig` (2018) and `speculator.py.orig` (768) -- 3,685 lines of
copied VLLM source that are nobody's patch and serve no purpose. Every install
drops that junk into the deployed vllm tree.

That is more than untidy. A new-file hunk collides with the `.orig` backup GNU
`patch` itself writes when an earlier patch applies with an offset: a plain
`patch -p1` run (the README and Dockerfile form, no
`--no-backup-if-mismatch`) then prompts

```
The next patch would create the file v1/worker/gpu/model_runner.py.orig,
which already exists!  Assume -R? [n]
```

skips the hunk and exits non-zero, killing the series mid-apply. A pristine
wheel with no offsets probably never trips the collision, but the junk files
get installed either way, and the collision is one drift away from a red build.

## Change

Drop only the three backup-file hunks. Nothing else moves: the commit is 3,703
pure deletions, and no other hunk content in either file is touched.

## Verification

- The patch-integrity script is green on this branch against a pristine vLLM
  0.28.0 checkout: full series in GNU patch, DFlash2 contract order in
  `git apply --whitespace=error`.
- The README/Dockerfile plain `patch -p1` form now applies all 30 patches with
  no "already exists" prompt (before, it hit one at
  `dflash2-lookup-drafting.patch` on an offset tree).

## Note for already-deployed installs

Installs made with the old patches still carry the `.orig` files; reinstalling
does not remove them. One-off cleanup:

```bash
find $(python -c 'import vllm, os; print(os.path.dirname(vllm.__file__))') -name '*.orig' -delete
```

## Relationship to #54 (series ordering)

Independent: this PR only deletes hunks, it does not touch names or order, so
it can land before or after #54 without a rebase.


mhenrichsen:
Merging. I re-ran the whole thing against a pristine v0.28.0 checkout and the deletion is exactly what it says: 3,703 lines removed, all of them inside three file-creation hunks for `.orig` backups, no other hunk content touched.

```
== pass 1: the whole series, GNU patch, glob order
   30 patches applied, 0 with an offset
patch integrity: OK
```

and with the series applied by hand, `find -name '*.orig'` in the vllm package goes from 3 files to 0.

**One thing this fixes that the PR does not claim.** `patches/_check_applied.py` builds its file list from every `+++` line in a patch, so the three `.orig` paths were *files verify.sh required to be present in the installed tree*. That makes the cleanup you document at the bottom of the PR body actively break verification. Reproduced on a patched tree:

```
dflash2-lookup-drafting: applied(0)
dflash2-prewarm: applied(0)
--- find -name '*.orig' -delete ---
dflash2-lookup-drafting: NOT applied(1)
dflash2-prewarm: NOT applied(1)
--- same tree, this branch's patch files ---
dflash2-lookup-drafting: applied(0)
dflash2-prewarm: applied(0)
```

That is the failure #90 reports on the native 3090 ("its verify.sh showed two failures at the time, since traced to deleted `.orig` backups that only the verifier reads"). It was not a stale venv, it was this. After this lands, a box that has run the cleanup verifies green.

The `patch -p1` collision you describe did not fire for me on a pristine tree with either form, which matches your own "probably never trips the collision" caveat. It is not why I am merging. The 3,685 lines of vLLM source installed into every tree, and the verifier false negative above, are.

Independent of #54, as you say, and I will renumber this when that lands rather than the other way round.

PR #93, opened 2026-09-10
alternative.sh: turn on the int4 multi-query 3D verify by default; wire INT8_ACT like start_qwen.sh; register the knob

cpuchip:
1. The kernel in `patches/spec-decode-int4-kv-mq3d.patch` was opt-in (`VLLM_INT4_MQ_3D`, default 0) and no launcher set it, so the int4 profile ran the stock 2D verify. This turns it on for `alternative.sh` (`INT4_MQ_3D=${INT4_MQ_3D:-1}`); `INT4_MQ_3D=0` restores the old behaviour.
2. Wires `INT8_ACT` / `INT8_LAYERS` the way `start_qwen.sh` already does (opt-in, off by default). It is a prefill knob: the paired 4090 repro measured decode 8 to 13 percent lower and the pool 22 percent smaller with it on, and the launcher header says so.
The switch is exported by `alternative.sh` only, because the int4 per-token-head cache is that launcher's profile; `start_qwen.sh` and `batch/start_qwen.sh` never run the int4 KV path, so the knob has nothing to switch there.
3. `patches/int4-mq3d-envs.patch` registers `VLLM_INT4_MQ_3D` and `VLLM_INT4_MQ_3D_DEBUG` in `envs.py` the way `speed-knobs-envs.patch` registers the other single-user knobs, so the switch is inside vLLM's registry and the torch.compile cache key now that it is a default (review ask). Series line at the tail. The kernel reads both knobs through `envs` now (`spec-decode-int4-kv-mq3d.patch` and the `spec-decode-scratch-token-units.patch` hunks around it are regenerated), so the registry is the only reader.

**Measured** on a 4090 (WSL2), DFlash2 k=7, fresh prefill, greedy, 256 out, median of 3, 120k model length: 2D verify 43.9 / 26.5 / 16.6 tok/s at 24k / 49k / 88k prompt tokens; 3D verify 90.4 / 73.7 / 60.2 (200k boot, 235,932-token pool), 48.1 at 145k. Short prompts (`bench/real_rep.sh`, eight prompts, 1,024 out): the crossover floor is -1% at default sampling and -3% greedy.

**Native 3090, captured path** (the maintainer's rows from review, now in `docs/spec-decode-scratch-token-units.md`): `MAX_LEN=98304 GPU_UTIL=0.93 PREFIX_CACHE=0 MAX_SEQS=1`, k=7, pool 185,373 on every boot, compile cache wiped per arm. Quality on 200 GSM8K rows: perplexity agrees to four significant figures on every split (all 8.2079 vs 8.2087), GSM8K 0.945 vs 0.955 (two questions, inside n=200). Decode, prefill excluded: 1.72x / 2.12x / 2.42x at 24k / 49k / 88k on a repeat prompt and 1.78x / 1.33x / 2.50x on wikitext prose; prefill untouched (worst gap 1.35%). So the range on a 3090 is 1.3x to 2.5x; the larger multiple the first version of this body quoted was the WSL2 4090 row at 88k and does not reproduce in magnitude on the 3090. The arms diverge at T=0 on long prose prompts: the 3D reducer is a different reduction order, and this flip does not claim bit-for-bit equal text between the arms.

**What "3D on" covers at k=7 and the defaults** (`MAX_SEQS=1`, `DFLASH_TOKENS=7`): the launcher sets `max_cudagraph_capture_size = MAX_SEQS * (DFLASH_TOKENS + 1) = 8`, so the captured decode shapes are at most 8 query tokens wide; the 9-wide verify (8 plus the bonus position) is not a captured shape at the defaults and runs outside the captured graphs. The earlier version of this body listed 9 among the captured shapes; that was wrong for the defaults and is corrected here. Prefill chunks (up to 2048 query tokens) never take the 3D path.

**Verified on the native 3090 against this head, in an image built from the branch** (so the new reader is the installed one): series integrity 33 applied, 0 rejected; the registry reads unset as False and "1" as True (the vLLM default stays off; "on by default" is the launcher's `INT4_MQ_3D=${INT4_MQ_3D:-1}`, and this image run without `alternative.sh` keeps the 3D path off); the boot at the defaults captures `max_cudagraph_capture_size` 8 with sizes 1, 2, 4, 8, so the 9-wide verify runs outside the captured graphs as the body says; and the sense test on served verifies, counted after graph capture: with the knob unset, `use_3d` is True at q=1, q=9 and q=16; with `INT4_MQ_3D=0`, q=1 stays True and q=9 and q=16 are False with the debug reason `mq_but_flag_off` counted exactly once per multi-query call (32 of 32), so the flag gates multi-query only, as named. One note for anyone comparing text between the arms: two boots at the same environment diverge at the same character as the two arms do, so a cross-boot text comparison cannot test this flag on that box; the counters are the instrument.

**Verified before filing:** `verify.sh --no-server` OK on the 4090 image; the series in `patches/series` order applies to a pristine v0.28.0 checkout with the new patch landing on exact context in its position (`check_vllm_series.sh`: 33 applied, 0 rejected). The oracle `bench/mq3d_layer2_oracle.py` is 10/10 on both boxes (eager path by its own config line).

**What is still unproven:** the launcher's real default `MAX_LEN=256000` and the 145k depth on the native card (the maintainer ran 98304 to keep both arms in one pool geometry); the short-prompt crossover on the native card; `INT8_ACT` on this profile natively; k above 7 and the lookup regime; whether registering the knob changes the compile-cache hash on this vLLM (it is registered and read as the precedent patch's knobs are; a warm-cache toggle A/B is the behavioural check and has not been run); GSM8K at n=200 cannot resolve a one-point gap.


kucheryavenkovn:
Independent reproduction on RTX 4090 24 GB / Windows 11 / WSL2 — PR reproduced, numbers confirmed and exceeded.

## Setup
- Image `ghcr.io/syv-ai/qwen38-27b-rtx3090:sha-0e95195` (vLLM 0.28.0), driver 616.56, Docker 29.7.2, WSL2 2.7.13
- Qwen3.8-27B-W4A16-AutoRound + DFlash2-W4A16 drafter, k=7 probabilistic, TRITON_ATTN + int4_per_token_head KV, pool 213,975 @ MAX_LEN=160k, GPU_UTIL=0.88, MAX_SEQS=1
- Protocol: greedy (T=0), 1500 output tokens with ignore_eos, unique prompt per run (fresh prefill, cached_tokens=0 everywhere), the SAME seed set paired across arms, spec-decode + preemption counters snapshotted per request, WDDM shared-memory sampled at 5 Hz from Windows
- The kernel itself (PR #46 + #57/#67) is already in the image, so A/B is exactly `VLLM_INT4_MQ_3D=0` vs `=1`, nothing else. Boot facts identical in both arms: FULL_AND_PIECEWISE, capture sizes [1,2,4,8], same KV pool.

## Runtime wiring check (VLLM_INT4_MQ_3D_DEBUG=1, debug boots only)
- `=1`: verify rows `tokens=8 num_seqs=1 max_seqlen_q=8 use_3d=True | reasons=NONE(3d eligible)` (also 9-row tails); prefill (q=1696) correctly stays on the 2D path (`max_query_len_policy>16`)
- `=0`: same verify rows `use_3d=False | reasons=mq_but_flag_off` — old path confirmed
- breach_count=0 / degraded=0 in every boot

## Table 1 — 2D vs 3D, median decode tok/s (3 runs per depth, 5 at 131k)

| Context | 2D | 3D | Speedup | tok/step | Acceptance | Prefill 3D, tok/s | TTFT 3D, s | Spill |
|---|---|---|---|---|---|---|---|---|
| 24k | 64.4 | 128.1 | 1.99x | 3.76 | 0.394 | 1915 | 12.6 | no |
| 49k | 39.2 | 105.7 | 2.70x | 3.66 | 0.380 | 1461 | 33.6 | no |
| 88k | 23.9 | 90.5 | 3.80x | 3.80 | 0.401 | 1055 | 83.5 | no |
| 110k | 20.1 | 79.1 | 3.93x | 3.63 | 0.376 | 919 | 119.8 | no |
| 125k | 18.1 | 75.3 | 4.17x | 3.67 | 0.382 | 841 | 148.8 | no |
| 131k | 17.2 | 71.8 | 4.17x | 3.58 | 0.368 | 813 | 161.2 | no |
| 145k | 16.6 | 68.8 | 4.15x | 3.63 | 0.375 | 755 | 192.2 | no |

- Acceptance and tok/step are statistically identical across arms (paired seeds), so the gain is purely the verify kernel — the mechanism matches the PR description
- Prefill/TTFT unchanged between arms, as claimed
- A 131k recheck on a fresh boot (same seeds) gave 70.7/67.6 — within the run spread, no thermal drift
- Spill-clean: WDDM shared 200–261 MB during all measured runs (baseline ~200 MB); 0 preemptions everywhere
- Absolute numbers are higher than the ones in the PR comment (better acceptance on these prompts → more tok/step), but the per-depth speedup ratios match well: 2.0x/2.8x/3.8x here vs 2.1x/2.8x/3.6x (24k/49k/88k) in the PR

## Table 2 — INT8_ACT=int8 on the 3D arm (VLLM_MARLIN_INPUT_DTYPE=int8, INCLUDE_RE mlp|linear_attn|self_attn)

| Context | 3D decode | 3D+INT8 decode | Prefill gain | Decode change | KV pool |
|---|---|---|---|---|---|
| 24k | 128.1 (pf 1915) | 110.7 (pf 2687) | +40% | -13.5% | 213,975 |
| 88k | 90.5 (pf 1055) | 79.8 (pf 1218) | +15% | -11.9% | → |
| 131k | 71.8 (pf 813) | 66.0 (pf 891) | +10% | -8.1% | 167,068 |

Prefill claim confirmed (+40% at 24k vs +42% in the PR). "Decode unchanged" did NOT reproduce here: steps/s drops ~5-6% plus ~2% acceptance, net -8..-13% decode, and the W4A8 path costs 46.9k pool tokens (-22%). Worth a caveat in the docs.

## Table 3 — reference: MTP + FP8 + FlashInfer vs DFlash2 int4 3D (same bench, same seeds)

| Context | DFlash2 int4 3D | MTP FP8 FlashInfer | Winner |
|---|---|---|---|
| 88k | 90.5 | 33.1 | int4 2.7x |
| 110k | 79.1 | 47.0 | int4 1.7x |
| 131k | 71.8 | 33.6 | int4 2.1x |

FP8 caveats: that profile runs GPU_UTIL=0.93 → WDDM shared 580–650 MB during runs, 1–2 preemptions per request, and one 131k run (16.9 tok/s) was spill-contaminated (700–825 MB shared) and excluded. FP8 decode 33–35 tok/s matches our earlier session, so it reproduces; it is just far behind at 131k, and its prefill is 1.7–2.7x slower too.

## Verdict
1. PR works as described on RTX 4090/WSL2; the default-on flip is safe and correct.
2. At 131k: 71.8 tok/s (3D) vs 17.2 (2D) = 4.17x.
3. INT8_ACT=int8: take it only for prefill-heavy workloads.
4. At ~131k on this card, DFlash2+int4+3D beats MTP+FP8+FlashInfer by ~2x with more VRAM headroom.

Full per-run data: 66 runs (per-request JSON with spec-decode deltas, TTFT, prefill/decode rates + a 5 Hz WDDM shared-memory log) — happy to share/push the CSV if useful.

mhenrichsen:
The throughput case is made and I am not going to ask you to re-measure it. Two cards, prefill identical at every depth in the 3090 control, short prompts checked for a shallow cost — that is the right shape for a default flip. What is missing is the other half, and the repo already wrote down what it is.

`docs/MR-DRAFT.md` says this kernel ships default-off for a stated reason:

> It ships opt in (`VLLM_INT4_MQ_3D=1`, default off) because six checks are still owed before anyone should trust it as a default: a 72k operator/logit oracle, full length exactness, per position logit comparison, eager+captured+q=9 parity, a shallow crossover floor (shallow currently reads -16%), and capture aware dual variant switching.

This PR flips that default and does not mention the list. Some of it has since been discharged and the PR should say so rather than leave a reader with a doc that contradicts the launcher:

- **Shallow crossover** — discharged here. Your `bench/real_rep.sh` rows replace the -16% with level. One ask: say which column is which. "130.2 vs 131.6 tok/s default sampling, 150.6 vs 155.3 greedy" reads as 3D being the lower number in both pairs, and if that is right then "level" is a -1% / -3% short-prompt cost bought with a 3.6x long-prompt gain, which is a fine trade but a different sentence.
- **Operator oracle and per-position comparison** — largely discharged by `bench/mq3d_layer2_oracle.py` and its verdict records (`docs/spec-decode-scratch-token-units.md`), which run both legs against an fp32 reference over its own dequantisation, with `_launch_packed_attn` call-counted so a green that never reached the production path cannot occur. Cite it. That is the strongest evidence this kernel has and it is not in the PR.
- **eager + captured + q=9 parity, and capture-aware dual-variant switching** — not discharged, and they are the two that your change actually exposes, because flipping the default is what puts the 3D path inside the captured decode graphs for every user of this profile. The oracle above declares eager, no CUDA graphs. Your PR asserts the capture shapes 1/2/4/6/8/9 all take the 3D path; asserting which path runs is not the same as showing the two paths agree there.

So the concrete ask is small: **one quality row on the captured path at this profile's default**, `INT4_MQ_3D=0` against `INT4_MQ_3D=1`, same seeds, on either card. `bench/quality_battery.py` with `PREFIX_CACHE=0` is enough. The depth-15 quality rows in `docs/spec-decode-scratch-token-units.md` are the only ones that exist today and they move around (judgment 8/8 stock, 6/8 with the 3D path off, 6/8 with it on), which is noise at that sample size rather than a clean bill.

I am asking because I cannot run it. We have no GPU here at the moment, so for anything whose evidence is runtime behaviour I am reduced to reading your numbers and checking they are self-consistent — and on a correctness-adjacent default flip, throughput numbers alone are not the thing to be convinced by. Everything I *could* check is fine: `alternative.sh` has `set -e` at line 11 and your `[ -n "$INT8_ACT" ] && export ...` lines are the last command in neither position that would trip it, matching what `start_qwen.sh` already does.

Two smaller things, both one-line:

- `docs/MR-DRAFT.md` line 70 becomes false the moment this merges. Update it in this PR.
- `VLLM_INT4_MQ_3D` is read with `os.environ.get` rather than being registered in vLLM's `envs.py`, so it is not in the torch.compile cache key. It was opt-in before and nobody swept it, so this never mattered; with it on by default, an A/B across the two legs on a warm cache is a stale-graph hazard of the kind `patches/speed-knobs-envs.patch` exists to prevent. Not a blocker for this PR, but if you run the quality row above, run each arm on a cold cache.

cpuchip:
Fair asks, and the doc-vs-launcher contradiction is the right thing to hold on. Taking the list in `docs/MR-DRAFT.md` in order, with what is discharged and where the evidence is.

**Shallow crossover**: discharged, and you read the columns right. 3D is the first number in each pair: default sampling 130.2 (3D) against 131.6 (2D), greedy 150.6 against 155.3. So it is a -1% / -3% short-prompt cost bought with the deep-end gain, not a wash. I will say it that way in the PR rather than "level".

**Operator oracle and per-position comparison**: discharged by `bench/mq3d_layer2_oracle.py` and the verdict records in `docs/spec-decode-scratch-token-units.md`: both legs against an fp32 reference over its own dequantisation, `_launch_packed_attn` call-counted so a green that never reached the production path cannot occur. Citing it in the PR; agreed it is the strongest evidence here.

**eager + captured + q=9 parity, and capture-aware dual-variant switching**: this is the one that needed a run, and here it is. Captured path, `INT4_MQ_3D=0` vs `=1`, same 200 GSM8K rows at T=0, cold compile cache each arm, `PREFIX_CACHE=0`:

| arm | GSM8K acc (n=200) | mean tokens | perplexity (all / en / da) |
|---|---|---|---|
| 2D (INT4_MQ_3D=0) | 0.950 | 370 | 10.8102 / 10.7296 / 10.8754 |
| 3D (INT4_MQ_3D=1) | 0.950 | 375 | 10.8135 / 10.7326 / 10.8790 |

Identical accuracy, perplexity matching to three significant figures. So the two paths agree on the captured graphs the default flip exposes, not just "which path ran". This replaces the depth-15 rows that moved around; at n=200 the noise you flagged is gone.

The two one-liners: `docs/MR-DRAFT.md` line 70 is updated in this PR to mark shallow-crossover, the oracle, and captured parity discharged and leave the native sm89 row as the open item. And you are right that `VLLM_INT4_MQ_3D` is read with `os.environ.get` and is not in the compile-cache key; the A/B above was run cold-cache per arm for exactly that reason, and registering it in `envs.py` is worth doing as a follow-up so a warm-cache A/B is not a stale-graph trap.

cpuchip:
Thank you, this is a thorough reproduction and the paired-seed protocol is exactly right. Two things I want to accept into the record.

Your 4.17x at 131k (71.8 vs 17.2) and the flat acceptance across arms confirm the mechanism is purely the verify kernel, which is the claim. Good to have it independently at higher absolute numbers than mine.

On INT8_ACT: your paired-seed measurement (steps/s down ~5-6%, acceptance down ~2%, net -8 to -13% decode, and -22% pool) is more careful than mine, and it corrects the "decode unchanged" line in the PR. My own arms were noisier and I read them as flat; yours are the better data. So the honest guidance is INT8_ACT is a prefill-only lever (the +40% at 24k holds) and it costs decode and pool, not free. I will strike the "decode unchanged" wording rather than leave a claim your data contradicts. Please do share the CSV; the 5 Hz WDDM shared-memory log especially is worth keeping beside these rows.

mhenrichsen:
Holding this one comment short of merge. The kernel flip itself is confirmed on a real 3090 — it does what you say it does — but two one-line items in the branch have to change first, and both are quick.

**What I ran.** Production stopped, card idle, `~/q28` with a copy of this branch's `alternative.sh`, `MAX_LEN=98304 GPU_UTIL=0.93 PREFIX_CACHE=0 MAX_SEQS=1`, `SPEC=dflash2` k=7, power limit untouched at 250 W, torch.compile cache wiped before each boot. All three boots landed on the same pool (`GPU KV cache size: 185,373 tokens`, concurrency 1.89x), so the arms are comparable. Wiring verified out of `/proc/<pid>/environ`: arm A `VLLM_INT4_MQ_3D=0`, arm B (nothing set in the shell) `VLLM_INT4_MQ_3D=1` — the default flip is live, and the kernel is in the install (`int4_per_token_head.py:779`).

Quality, `bench/quality_battery.py` from this branch, 200 GSM8K rows, greedy, thinking off:

```
2D  PPL all 8.2079   GSM8K n=200 acc=0.945  mean_tokens=381
3D  PPL all 8.2087   GSM8K n=200 acc=0.955  mean_tokens=375
```

Perplexity agrees to four significant figures on every split (en -0.05%, da +0.02%, code +0.08%, all +0.010%); GSM8K is 1.0 absolute point apart, i.e. two questions. Both inside your criteria. Worth noting this is the captured-graph path (`max_cudagraph_capture_size=8`, k=7) — precisely the row `bench/mq3d_layer2_oracle.py` does not cover, since its own config line says `cudagraph: NONE (eager)`. So your 4090 captured-path numbers now have a native-3090 companion.

Decode, isolated as `255/(t_256 - t_1tok)` on a prefix-cached prompt so prefill is excluded (your plan's `completion/wall` formula silently measures a cached prefill — 0.28 enables prefix caching by default and `PREFIX_CACHE=0` only omits the flag, it does not pass `--no-enable-prefix-caching`; that's how you get 2.5e11 tok/s):

```
degenerate repeat prompt        2D      3D
  24k   decode tok/s         119.09  205.34   1.72x
  49k                         75.21  159.79   2.12x
  88k                         19.04   46.06   2.42x

wikitext-2 prose (added by me)
  24k                         44.17   78.80   1.78x
  49k                         35.12   46.83   1.33x
  88k                         15.18   38.02   2.50x
```

Prefill is untouched at every depth in both prompt sets (worst gap 1.35%, prompt_tokens identical per depth) — the change is verify-only as advertised. I added the prose prompt because the repeat-the-fox prompt inflates DFlash2 acceptance and could have carried the whole win by itself; it didn't, 88k still doubles. But the honest range on this card is **1.3x–2.5x, not 3.6x**; that headline is a 4090/WSL2 number and it does not reproduce in magnitude here.

One thing to record: the arms diverge at T=0 on the long prose prompts (49k: "...was a major facto" vs "...was a great boost"; 88k: "He was 20 years old , 106" vs "He was 19 years and 77 day"; 24k identical). Aggregate quality doesn't move and the repo already documents that greedy-with-speculation isn't bit-exact, but the 3D reducer is a different reduction order, so "identical output" isn't a claim this flip gets to make. Please don't put it in the docs.

**Two blockers, both one-liners:**

1. `single-user/alternative.sh:32-33` still carries "...835 tok/s against 1,942 / 1,446 / 1,039 / 736 on a 4090, 200k max length) **with decode unchanged**". You struck that clause from the PR body on 09-11 after the 4090 repro measured -8..-13% decode and -22% pool. Merging as-is ships a retracted claim as an inline comment for the next reader. Replace it with the prefill-only wording you already agreed to.

2. `VLLM_INT4_MQ_3D` is read with `os.environ.get` inside the kernel (`patches/spec-decode-int4-kv-mq3d.patch:26`, `spec-decode-scratch-token-units.patch:272`) and is not in vLLM's `envs.py` — I checked the install, only `VLLM_MARLIN_INPUT_DTYPE` is registered. That puts it outside the torch.compile cache key: harmless while opt-in, a stale-graph hazard the moment it's the default and someone A/Bs it warm. I dodged it by wiping `~/.cache/vllm/torch_compile_cache` before every boot; the next person won't. Either register it via a patch (`patches/speed-knobs-envs.patch` is the precedent) or add a gotcha saying a warm-cache A/B of this flag is invalid. I'd take the patch.

**Two smaller asks:** paste the captured-path GSM8K/ppl A/B into `docs/spec-decode-scratch-token-units.md` so the evidence for the default lives in the repo rather than only in this thread — my rows above are yours to use — and confirm the q=9 claim against this launcher's capture list, since `CG = MAX_SEQS*(DFLASH_TOKENS+1) = 8` at the defaults, so a 9-token verify isn't a captured shape.

**What I did not establish**, so nobody reads more into the above than is there: the launcher's real default `MAX_LEN=256000` and the 145k depth (I ran 98304 to keep both arms in one pool geometry and inside the time budget), the short-prompt crossover on `bench/real_rep.sh`, INT8_ACT on this profile, k>7 or the lookup regime, the sm89 row the MR-DRAFT still lists open, and blocker (2) itself — that one is argued from the source, not measured. GSM8K at n=200 also cannot resolve a 1-point gap as signal or noise.

Credit where it's due: the `docs/MR-DRAFT.md:70` rewrite in 3d2320a discharges the line-70 ask properly, the oracle harness (`bench/mq3d_layer2_oracle.py` + the two verdict jsonls, call-counted) is real work and made this cheap to check, and the launcher wiring is byte-identical in shape to `start_qwen.sh:104-107` as claimed. Fix (1) and (2), push, and I'll merge without re-running the card.

Production is back up: `qwen-single` active with both drop-ins, /health 200, 250 W. The deployed `alternative.sh` is still main's — my copies and probe scripts are off the box.

@kucheryavenkovn — nothing blocking from my side, and yes, the CSV is worth taking.

cpuchip:
Pushed, both blockers and both asks:

1. The `alternative.sh` header no longer says decode was unchanged by `INT8_ACT`; it says what was measured (decode 8 to 13 percent lower and the pool 22 percent smaller with it on, on the 4090 repro), so the retracted clause is gone from the inline comment as well as the body.
2. `VLLM_INT4_MQ_3D` and `VLLM_INT4_MQ_3D_DEBUG` are registered in `envs.py` by `patches/int4-mq3d-envs.patch` (after `speed-knobs-envs.patch`, series line at the tail), and the kernel reads both through `envs` rather than `os.environ`, so the registry is the only reader and the knob sits in the compile cache key the way the precedent knobs do.
3. Your captured-path rows are in `docs/spec-decode-scratch-token-units.md`, with the 3090 range (1.3x to 2.5x) replacing the 4090 multiple in the doc and in `docs/MR-DRAFT.md`.
4. q=9: at the launcher's defaults `max_cudagraph_capture_size = MAX_SEQS * (DFLASH_TOKENS + 1) = 8`, and a boot at those defaults captures sizes 1, 2, 4 and 8, so the 9-wide verify is not a captured shape; the earlier body was wrong to list it, and it is corrected.

Verified on a native 3090 in an image built from the branch: series 33 applied, 0 rejected; the registry reads unset as False and 1 as True (the vLLM default stays off; on-by-default is the launcher's line); and the sense test on served verifies after graph capture: knob unset, `use_3d` True at q=1, 9 and 16; `INT4_MQ_3D=0`, q=1 True and q=9, q=16 False with `mq_but_flag_off` counted once per multi-query call (32 of 32). The T=0 divergence you saw between arms reproduces between two boots at the same environment, so text comparison cannot test the flag on that box; the counters can.

Still unproven: the 145k depth and the 256k default length natively; the short-prompt crossover natively; whether registration changes the compile-cache hash on this vLLM (argued from the precedent patch, not measured by a warm-cache toggle).

mhenrichsen:
Merged. Both blockers from the last round are discharged in the branch, and I checked them by reading the branch rather than taking the summary's word.

**Blocker 1 (retracted claim).** `single-user/alternative.sh` no longer carries "with decode unchanged". The header now states INT8_ACT as a prefill-only lever that costs 8-13% decode and 22% pool, which is what the 4090 repro measured. Correct.

**Blocker 2 (compile-cache key).** `patches/int4-mq3d-envs.patch` registers both knobs, the series lists it at the tail after `speed-knobs-envs.patch`, and both kernel patches now read `envs.VLLM_INT4_MQ_3D` / `envs.VLLM_INT4_MQ_3D_DEBUG`. The name resolves: after the full series, `_launch_packed_attn` starts at `int4_per_token_head.py:698`, its function-local `import vllm.envs as envs` is at 736, and the reads are at 779 and 854 — same function, import first, so no NameError. I also settled the one item you listed as argued-not-measured, from source and without a GPU: `vllm/envs.py:compile_factors()` hashes every registered env var except an explicit `ignored_factors` set, and `vllm/compilation/backends.py:1031-1066` folds that hash into the compile cache directory key. Neither new knob is in `ignored_factors`, so registration does exactly what the precedent patch's header claims and the warm-cache stale-graph hazard is closed, not dodged.

One thing I checked because it would have been silent: `bench/mq3d_layer2_oracle.py:301` sets `os.environ["VLLM_INT4_MQ_3D"]` at runtime, and an env cache would freeze that and turn the two legs into one. It does not happen here — `envs.__getattr__` re-evaluates its lambda per access until `enable_envs_cache()` runs, and that fires only from the engine core and the multiproc executor, neither of which the oracle builds. The oracle also asserts the per-leg reason bits, so a frozen knob would fail rather than pass both legs.

**Two chores I did myself rather than spend another round-trip on you.** Pushed to your branch before merging:

- `docs/MR-DRAFT.md:24-27` still described the kernel as "opt in ... defaults off ... tests it still owes", contradicting your own rewrite of section 3 in the same PR. Same class of defect as the line-70 ask, just in the summary list. It now says on by default in `single-user/alternative.sh` with `INT4_MQ_3D=0` as the way back, carries the 1.3x-2.5x native-3090 range, and marks the ~3.3x figure as the 4090/WSL2 row it is.
- New gotcha 56 in `docs/gotchas.md` records the compile-cache finding, because it has a live consequence for this repo's habits: `VLLM_INT4_MQ_3D_DEBUG=1` now changes the cache key, so a debug boot recompiles cold and must never be timed against a warm production boot. It also notes that the per-arm `rm -rf ~/.cache/vllm/torch_compile_cache` I did by hand in the last round is no longer required.

Main moved twice under the PR while I was doing this (#118 and the offload/SSE merges), so the branch carries two merge commits from me. Both conflicts were mechanical: `patches/series` keeps `int4-mq3d-envs.patch` at the tail after the newly landed entries, and my gotcha renumbered to 56 behind the SSE keep-alive one. Re-ran `bash patches/check_vllm_series.sh` against a pristine vLLM v0.28.0 checkout after each resolution — 37 patches applied, 0 with an offset, `git apply --check` clean on the five ordered DFlash patches, "patch integrity: OK". CI git-apply is green on the merged head.

To state plainly what my last round was right about and what it was not: a default flip did need the captured path measured, and it has been — `max_cudagraph_capture_size=8` at k=7, same pool (185,373 tokens) on all boots, cache wiped per arm, wiring read out of `/proc/<pid>/environ`, PPL 8.2079 vs 8.2087 and GSM8K 0.945 vs 0.955 at n=200, decode 1.3x-2.5x with prefill excluded. Those rows are in `docs/spec-decode-scratch-token-units.md` now, so the evidence for the default lives beside the default. The T=0 divergence between arms on long prose stands, and the PR is right not to claim bit-identical text — keep that wording as written. Note the MR-DRAFT section-3 parity row quotes your 4090 numbers (0.950 both, PPL 10.8102 / 10.8135) while the doc records my 3090 run; both are real runs from this thread, but if you touch that paragraph again, say which card each row is from.

Still open, and not a reason to hold a 3090 repo's 3090-measured default: a native sm89 row (every 4090 number here is WSL2), the 145k depth and the 256k default length natively, and the short-prompt crossover natively. If you get a native 4090 in front of you, the sm89 row is the one worth running — the same two arms, `bench/quality_battery.py` with `PREFIX_CACHE=0` and cold cache per arm, and it closes the last ledger line in section 3.

Credit where it belongs: the oracle harness with `_launch_packed_attn` call-counted is the strongest evidence this kernel has and is what made the review cheap, the envs patch is the right fix rather than a gotcha note, and you turned two review rounds around same-day.

PR #96, opened 2026-09-11
kv sharing: let the full-attention layers share a KV cache (experiment, needs a GPU to judge)

mhenrichsen:
Cross-layer KV cache sharing for the 16 full-attention layers, off by default.
**The quality cost is unmeasured** — I had no GPU while writing this — so this is
a request for measurement as much as a patch.

## Why

DeepSeek-V4.1-Flash reports 890 bytes per token of KV cache. A large part of how
it gets there is that most of its attention layers do not own a cache: across 40
layers only four compress their own and the rest read those. Our 16
full-attention layers each own one, at 64 KiB/token combined at fp16.

vLLM already implements the mechanism. An `Attention` layer built with
`kv_sharing_target_layer_name` is skipped when the runner builds cache specs, so
no memory is allocated for it, and its tensor is aliased to the target's. It
validates that the target precedes the layer and matches its attention type.
The Qwen3.5 model code just never passes the argument, so the capability was
unreachable. This passes it.

```
KV_SHARE=group:2    # 8 of 16 layers own a cache, global KV cache halved
KV_SHARE=group:4    # 4 owners, quartered — DeepSeek's owner count
KV_SHARE=suffix:4   # 4 owners, You Only Cache Once layout
```

Unset is stock behaviour and changes nothing.

`group:N` groups consecutive full-attention layers N at a time and gives the
cache to the first of each, so the target is always nearby. `suffix:K` keeps the
first K owners and points everything later at the K-th; it should be the more
damaging layout, and it is the only one `--kv-sharing-fast-prefill` can use,
since that flag needs the sharing layers to form a suffix.

`VLLM_QWEN_KV_SHARE` is registered in `envs.py` so it takes part in the
torch.compile cache key. It changes which layers own a cache and therefore the
compiled graph; without that, a sweep would replay the previous arm's graph.

## What is verified, without a GPU

- The series applies to a pristine vLLM 0.28.0: 31 patches, 0 offsets, both CI passes.
- The layout helper, exercised over all 64 layers: for `group:2`, `group:4`,
  `suffix:8` and `suffix:4` every target precedes its layer and is itself an
  owner, so there are no chains. `group:1`, `suffix:0`, `suffix:16` and junk are
  rejected with a message.
- The launcher exports the variable only when `KV_SHARE` is non-empty.
- The sweep's log parsing, against a realistic log line.

## What is not verified

Everything that needs the card. It has never booted. In particular I expect
trouble in **cache group geometry** before model quality: shared layers get
appended to their target's group, and group composition on this hybrid model is
where #57 and #63 both came from.

## The ask

`bash bench/kv_share_sweep.sh` runs the arms and scores them on perplexity,
GSM8K and needle retrieval at depth. It boots on port 18021 so it will not fight
a production server. Budget about 25 minutes per arm, or start with
`GSM_N=100 NEEDLE=0` for a first look.

A shorter first question, if the full sweep is too much: **does `KV_SHARE=group:2`
boot at all, and what does the log say about the KV pool?** That alone settles
the geometry risk, which is the part most likely to kill this.

The harness refuses three mistakes this repo has already made: it verifies each
arm took effect by reading the owner count out of the server log rather than
trusting the env var, it boots each arm twice so a cold compile cache is never
compared against a warm one, and it measures perplexity with `PREFIX_CACHE=0`.

Untrained, the honest expectation is `group:2` degrades and `group:4` falls
apart. The useful output is the shape of the curve, because that says whether a
light fine-tune on the attention layers could close the gap or whether these
layers are too dissimilar to share at all. If `group:2` is already broken, this
PR should be closed and the result written up as a gotcha — that is a perfectly
good outcome and it is cheaper than the alternative of finding out later.

`docs/kv-sharing.md` has the protocol, the ship/no-ship criteria, and the
arithmetic on why the other DeepSeek-flavoured idea doing the rounds right now
(Late Layer KV Approximation, the `Q3-8B-KVA-Projector` projector) does **not**
port to this model: it needs the back half of the network to be skippable, and
three quarters of our layers are gated-delta-net whose recurrent state cannot be
written per-position. Its ceiling here is 4% of prefill at 4k and 12% at 51k,
against the ~50% it gets on the all-attention Qwen3-8B it was built for.

Drive-by: the README docs table still said `gotchas.md` had 18 entries. It has 53.

### DIFF
--- a/README.md
+++ b/README.md
@@ -885,7 +885,7 @@ bug, `python bench/residue_sweep.py <tag>` (all 128 residues) with
 | | |
 |---|---|
 | [docs/optimizations.md](docs/optimizations.md) | Every optimization in full: why it was needed, what it measured, which patch implements it. Includes the two speculative-decoding modes (MTP and DFlash2) and the lookup drafter. |
-| [docs/gotchas.md](docs/gotchas.md) | 53 things that each cost us hours — read before debugging something that looks like a vLLM bug. |
+| [docs/gotchas.md](docs/gotchas.md) | 54 things that each cost us hours — read before debugging something that looks like a vLLM bug. |
 | [docs/quality.md](docs/quality.md) | IFBench, perplexity and GSM8K per configuration. |
 | [docs/docker.md](docs/docker.md) | The container image, and an independent WSL2 reproduction. |
 | [docs/kv-sharing.md](docs/kv-sharing.md) | Cross-layer KV cache sharing for the 16 full-attention layers: an unmeasured experiment (`KV_SHARE=group:2`) that would cut the global KV cache 2-4x if the quality holds, plus why the DeepSeek-V4.1-Flash prefill trick does not port to this model. |
--- a/README.md
+++ b/README.md
@@ -885,9 +885,10 @@ bug, `python bench/residue_sweep.py <tag>` (all 128 residues) with
 | | |
 |---|---|
 | [docs/optimizations.md](docs/optimizations.md) | Every optimization in full: why it was needed, what it measured, which patch implements it. Includes the two speculative-decoding modes (MTP and DFlash2) and the lookup drafter. |
-| [docs/gotchas.md](docs/gotchas.md) | 18 things that each cost us hours — read before debugging something that looks like a vLLM bug. |
+| [docs/gotchas.md](docs/gotchas.md) | 53 things that each cost us hours — read before debugging something that looks like a vLLM bug. |
 | [docs/quality.md](docs/quality.md) | IFBench, perplexity and GSM8K per configuration. |
 | [docs/docker.md](docs/docker.md) | The container image, and an independent WSL2 reproduction. |
+| [docs/kv-sharing.md](docs/kv-sharing.md) | Cross-layer KV cache sharing for the 16 full-attention layers: an unmeasured experiment (`KV_SHARE=group:2`) that would cut the global KV cache 2-4x if the quality holds, plus why the DeepSeek-V4.1-Flash prefill trick does not port to this model. |
 | [docs/long-context.md](docs/long-context.md) | 262k context with the KVarN 4/2-bit KV cache, what vLLM's own per-token-head KV modes are worth here, and how to run the DFlash2 drafter past 64k (`CTX=long`, 114-139k — worth it only for context reproduction). |
 | [batch/](batch/) · [single-user/](single-user/) | The two serving modes: full benchmark tables, every env knob, systemd units. |
 | [prepare/](prepare/) | The one-time model-preparation scripts run by [Setup](#setup) (and by `docker compose run --rm prepare`). |
--- a/bench/kv_share_sweep.sh
+++ b/bench/kv_share_sweep.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+# Cross-layer KV cache sharing sweep (patches/qwen3_5-kv-cache-sharing.patch).
+#
+# Asks the only question that matters about the feature: these 16 full_attention
+# layers were never trained to share a KV cache, so how much quality does each
+# sharing layout cost? The memory saving is arithmetic and needs no experiment --
+# group:2 halves the global KV cache, group:4 quarters it. The damage does.
+#
+#   bash bench/kv_share_sweep.sh                      # off group:2 group:4
+#   ARMS="off group:2 group:4 suffix:8" bash bench/kv_share_sweep.sh
+#   GSM_N=100 NEEDLE=0 bash bench/kv_share_sweep.sh   # quicker first look
+#
+# Results under bench/results-kv-share/<arm>/, one ROW line per measurement.
+# Budget roughly 25 minutes per arm with the defaults.
+#
+# THREE THINGS THIS SCRIPT REFUSES TO GET WRONG, because each of them has
+# produced a confident wrong answer in this repo before:
+#
+#  1. It verifies the arm actually took effect, by reading the owner count out
+#     of the server log and comparing it to the arithmetic. An arm that silently
+#     ran stock would report "no quality loss" and be believed. That is exactly
+#     how the patch-integrity CI job stayed green for months while checking
+#     nothing (docs/gotchas.md, the `git apply` prefix entry).
+#  2. It boots each arm TWICE and measures the second boot. VLLM_QWEN_KV_SHARE
+#     is part of the torch.compile cache key -- it has to be, it changes which
+#     layers own a cache -- so every arm's first boot compiles cold, and a cold
+#     compile profiles ~0.9 GiB more peak activation than a warm one. Comparing
+#     a cold arm against a warm arm is comparing compile-cache states, not
+#     layouts. Never A/B a KV pool across boots without this.
+#  3. It runs perplexity with PREFIX_CACHE=0. prompt_logprobs is corrupted by
+#     prefix caching on this hybrid model and reads ~23% high (gotcha 51).
+set -u
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO="$(dirname "$HERE")"
+cd "$REPO"
+ARMS=${ARMS:-"off group:2 group:4"}
+PORT=${PORT:-18021}          # do not fight the production port
+GSM_N=${GSM_N:-200}
+NEEDLE=${NEEDLE:-1}
+NEEDLE_TOKENS=${NEEDLE_TOKENS:-32000}
+ROOT="$HERE/results-kv-share"; mkdir -p "$ROOT"
+export PATH="$REPO/venv/bin:$PATH"
+export VLLM_API="http://127.0.0.1:$PORT/v1"
+SUMMARY="$ROOT/summary.txt"; : > "$SUMMARY"
+
+# 16 full_attention layers on Qwen3.8-27B (64 layers, full_attention_interval 4).
+FULL=16
+owners_for() {   # arm -> how many layers should still own a KV cache
+  case "$1" in
+    off)      echo $FULL ;;
+    group:*)  python3 -c "import math;print(math.ceil($FULL/${1#group:}))" ;;
+    suffix:*) echo "${1#suffix:}" ;;
+    *)        echo "unknown arm $1" >&2; exit 2 ;;
+  esac
+}
+
+stop_server() {
+  pkill -f "[v]llm serve" 2>/dev/null
+  for i in $(seq 1 60); do pgrep -f "[v]llm serve" >/dev/null || break; sleep 2; done
+  pgrep -f "[v]llm serve" >/dev/null && pkill -9 -f "[v]llm serve"
+  sleep 5
+}
+
+boot() {   # boot(arm, logfile) -> waits for health, returns 1 if it never came up
+  local arm=$1 log=$2
+  local share=""; [ "$arm" != off ] && share=$arm
+  KV_SHARE="$share" SPEC=${SPEC:-dflash2} CTX=${CTX:-fast} \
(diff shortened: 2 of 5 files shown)


mhenrichsen:
@cpuchip @mjungnickel18 — asking you two directly because this needs a card and I
do not have one for now. No obligation, and a "not this week" is a fine answer.

The patch makes the 16 full-attention layers able to share a KV cache, which
would cut the global KV cache 2x at `group:2` and 4x at `group:4`. The mechanism
is already in vLLM 0.28.0; the Qwen3.5 model code just never passed the
argument. So the engineering here is small and the only real question is what it
costs in quality, since these layers were never trained to share.

**The cheap version of the ask**, maybe ten minutes:

```bash
KV_SHARE=group:2 PORT=18021 bash single-user/start_qwen.sh
```

Does it boot, and what does the log say for `GPU KV cache size` against your
normal boot? There should also be one line reading `Qwen3.5 KV sharing
(group:2): 8 of 16 full_attention layers own a cache`. If that line is missing
the patch did not take and nothing else is worth running.

I think this is where it breaks. Shared layers get appended to their target's KV
cache group, and group composition on this hybrid model is exactly what produced
#57 and #63. Quality may never get a turn.

**The full version**, if the boot is clean and you have the time:

```bash
GSM_N=100 NEEDLE=0 bash bench/kv_share_sweep.sh     # ~10 min/arm, 3 arms
bash bench/kv_share_sweep.sh                        # ~25 min/arm
```

@cpuchip, the native 3090 is the reference configuration, so your number is the
one that would decide it — though you already have four PRs open here and I do
not want to add to that queue. @mjungnickel18, if WSL2 is what you have, the
quality arms travel fine even though the pool-size column will not; gotcha 53's
ceiling makes that half unreadable on WSL2 and the harness will still print it.

To be clear about what a useful result looks like: **if `group:2` is already
broken, that closes the PR and becomes a gotcha, and that is a good outcome.**
It is a much cheaper way to find out than discovering it after someone spends a
fine-tune on the idea. Do not feel you have to bring back good news.

Protocol and the ship/no-ship criteria are in `docs/kv-sharing.md`.

Dewamade:
@mhenrichsen 
RTX 3090, Ubuntu 24.04

GSM_N=100 NEEDLE=0 bash bench/kv_share_sweep.sh
=== arm off  (expect 16 of 16 full_attention layers to own a cache)
ROW off owners=16/16 pool=57669 tokens vram=23500 MiB, 250.00 W
kvshare-off PPL {'en': (10.7625, 11175), 'da': (10.9105, 13917), 'code': (3.3142, 7587), 'all': (8.2353, 32679)} 35s
kvshare-off GSM8K n=100 acc=0.940 mean_tokens=258 100s
ROW off kvshare-off PPL {'en': (10.7625, 11175), 'da': (10.9105, 13917), 'code': (3.3142, 7587), 'all': (8.2353, 32679)} 35s
ROW off kvshare-off GSM8K n=100 acc=0.940 mean_tokens=258 100s
=== arm group:2  (expect 8 of 16 full_attention layers to own a cache)
ROW group:2 owners=8/16 pool=87499 tokens vram=23500 MiB, 250.00 W
kvshare-group-2 PPL {'en': (16.1692, 11175), 'da': (14.0578, 13917), 'code': (5.6897, 7587), 'all': (11.9536, 32679)} 35s
kvshare-group-2 GSM8K n=100 acc=0.640 mean_tokens=371 152s
ROW group:2 kvshare-group-2 PPL {'en': (16.1692, 11175), 'da': (14.0578, 13917), 'code': (5.6897, 7587), 'all': (11.9536, 32679)} 35s
ROW group:2 kvshare-group-2 GSM8K n=100 acc=0.640 mean_tokens=371 152s
=== arm group:4  (expect 4 of 16 full_attention layers to own a cache)
ROW group:4 owners=4/16 pool=118188 tokens vram=23502 MiB, 250.00 W
kvshare-group-4 PPL {'en': (23.464, 11175), 'da': (21.3756, 13917), 'code': (12.7323, 7587), 'all': (19.5669, 32679)} 35s
kvshare-group-4 GSM8K n=100 acc=0.040 mean_tokens=698 280s
ROW group:4 kvshare-group-4 PPL {'en': (23.464, 11175), 'da': (21.3756, 13917), 'code': (12.7323, 7587), 'all': (19.5669, 32679)} 35s
ROW group:4 kvshare-group-4 GSM8K n=100 acc=0.040 mean_tokens=698 280s

=== summary ===
ROW off owners=16/16 pool=57669 tokens vram=23500 MiB, 250.00 W
ROW off kvshare-off PPL {'en': (10.7625, 11175), 'da': (10.9105, 13917), 'code': (3.3142, 7587), 'all': (8.2353, 32679)} 35s
ROW off kvshare-off GSM8K n=100 acc=0.940 mean_tokens=258 100s
ROW group:2 owners=8/16 pool=87499 tokens vram=23500 MiB, 250.00 W
ROW group:2 kvshare-group-2 PPL {'en': (16.1692, 11175), 'da': (14.0578, 13917), 'code': (5.6897, 7587), 'all': (11.9536, 32679)} 35s
ROW group:2 kvshare-group-2 GSM8K n=100 acc=0.640 mean_tokens=371 152s
ROW group:4 owners=4/16 pool=118188 tokens vram=23502 MiB, 250.00 W
ROW group:4 kvshare-group-4 PPL {'en': (23.464, 11175), 'da': (21.3756, 13917), 'code': (12.7323, 7587), 'all': (19.5669, 32679)} 35s
ROW group:4 kvshare-group-4 GSM8K n=100 acc=0.040 mean_tokens=698 280s

Baseline is the 'off' arm measured on the same box in the same sweep.
Numbers from a different session are not a baseline; boot-to-boot drift on
this stack is larger than several of the effects people have tried to claim.

cpuchip:
Booted and swept it on two boxes. Same picture as Dewamade's: the pool grows exactly as the arithmetic says and the quality goes with it.

**4090 under WSL2** (fork image at main, `bench/kv_share_sweep.sh` with `GSM_N=100 NEEDLE=0`, every arm booted twice and the warm boot measured, `PREFIX_CACHE=0` for the perplexity pass, `CTX=fast` so the KV pool is the launcher's pinned 5.2 GiB in every arm):

| arm | layers owning a cache | GPU KV pool | perplexity all (en / da / code) | GSM8K n=100 |
|---|---|---|---|---|
| off | 16 of 16 | 57,669 tokens | 8.23 (10.77 / 10.91 / 3.31) | 0.95 |
| group:2 | 8 of 16 | 89,620 | 11.95 (16.16 / 14.06 / 5.69) | 0.59 |
| group:4 | 4 of 16 | 121,891 | 19.58 (23.45 / 21.39 / 12.76) | 0.03 |

group:2 lands where Dewamade's 3090 did (his 11.95 and 0.64 against my 11.95 and 0.59, n=100 here). group:4 is not a usable model: perplexity more than doubles and GSM8K collapses to 3 of 100.

**Native 4090** (a rented bare-metal box, this image at `latest`, Ubuntu 24.04, driver 580.159, same sweep and knobs, the stock arm at the launcher's default length):

| arm | layers owning a cache | GPU KV pool | perplexity all (en / da / code) | GSM8K n=100 |
|---|---|---|---|---|
| off | 16 of 16 | 57,669 tokens | 8.23 (10.77 / 10.91 / 3.31) | 0.96 |
| group:2 | 8 of 16 | 87,499 | 11.95 (16.16 / 14.06 / 5.69) | 0.56 |
| group:4 | 4 of 16 | 118,188 | 19.57 (23.46 / 21.39 / 12.75) | 0.01 |

Native and WSL2 agree to the third digit on perplexity, and the group:2 pool on native is Dewamade's number exactly (87,499). So the WSL2 rows are not a WSL2 artefact.

**Native 3090** (an aged install, 0.28.0 plus the fork's patch series rather than the image; boot facts only, no quality pass on that venv): boots healthy in 135 s, the log line is there verbatim ("Qwen3.5 KV sharing (group:2): 8 of 16 full_attention layers own a cache"), GPU KV cache 73,246 tokens off against 125,145 with group:2, on 5.33 and 5.34 GiB of available KV (the same to within 0.01 GiB), so tokens per GiB 13,742 to 23,435. The pool grows because the per-token cache shrank, not because memory moved.

Two notes on the patch itself. `envs.py` hunk 2 lands at an 8-line offset on both the fork image and a from-scratch venv, so that is drift between the image's `envs.py` and pristine 0.28.0, not a box; a 0-offset CI run on pristine 0.28.0 is consistent with it. And the sweep's stock arm cannot boot at `MAX_LEN=60000` on `CTX=fast` (the unshared cache needs 5.3 GiB against the 5.2 GiB pin, the sharing arms fit), so the off row above is at the launcher's default 57,344; the pool count is pinned by `KV_MEM` and does not move with `MAX_LEN`, so the rows compare.

So the feature does what it says to memory, and group:2 costs roughly a third of the model's GSM8K accuracy for a 1.55x pool. That is the trade; whether it is worth an opt-in flag is your call.

cpuchip:
One correction to the native 3090 paragraph, caught by the person who ran it: that boot used the launcher's default drafter (MTP, k=4) with the pool unpinned, not the DFlash2 k=15 of the 4090 rows. So its 1.71x pool ratio (73,246 to 125,145) was never comparable to the 4090 rows' 1.52x to 1.55x, and the gap is expected rather than variance in the feature: the DFlash2 drafter carries its own KV cache that the 16-layer split never touches, and that unshared term dilutes the ratio.

Checked on the WSL2 4090 with the same sweep script and only the drafter changed (`SPEC=mtp`, k=4, `MAX_LEN=48000` so the stock arm clears the boot gate): off 57,600 tokens to group:2 95,675, a 1.66x ratio, against 1.55x with DFlash2 k=15 on the same card. The ratio moved most of the way toward the 3090's 1.71x when the drafter matched; the rest is the pinned-versus-floating pool budget. So the number to compare across boxes is each arm's pool at a stated drafter, not the ratio.

mhenrichsen:
Closing this. The PR pre-registered the criterion — "if `group:2` is already broken, close it and write it up as a gotcha" — and `group:2` is broken.

Three independent runs agree, on the reference configuration and on two other cards:

| arm | layers owning a cache | KV pool | PPL (all) | GSM8K n=100 |
|---|---|---|---|---|
| off | 16/16 | 57,669 | 8.24 | 0.94 |
| group:2 | 8/16 | 87,499 | 11.95 | 0.56–0.64 |
| group:4 | 4/16 | 118,188 | 19.57 | 0.01–0.04 |

`group:2` buys a 1.52× pool for ~36% relative GSM8K and +45% perplexity. `group:4` is not a model — one to four correct out of a hundred. That is the bottom rung of the ship/no-ship ladder in `docs/kv-sharing.md`, not the interesting middle, so there is nothing here worth carrying as a knob.

Thanks to @Dewamade and @cpuchip for spending the GPU time. Perplexity agrees to the third digit across a native 3090, a native 4090 and WSL2, and native-4090 `group:2` reproduced the pool size exactly at 87,499 — both guards held, the owner-count line printed verbatim in each arm and the off arm logged nothing, so these are real paired numbers and not a knob that silently did nothing.

One result worth keeping: the geometry risk I expected from the #57/#63 lineage did not materialise. It boots clean, and the pool grows because the per-token cache shrank at equal available-KV GiB. The failure is purely quality, not allocation.

Two defects in the harness, recorded in case anyone revives it:

1. `bench/kv_share_sweep.sh`'s `stop_server` runs an unfiltered `pkill -f "[v]llm serve"`. It boots on port 18021 specifically so it will not fight production — and then kills production on 18020 anyway. Fix that before running it on a live box.
2. The patch header, the launcher comment and the model-code comment all point at `bench/kv_share_sweep.py`; the file that ships is `.sh`.

The one hunk here that should survive is the README gotchas count (`docs/gotchas.md` ends at 54, README still says 18) — I will land that as its own commit along with the gotcha for this result.

Issue #95, opened 2026-09-10
0.28.0 kv_offload CPU tier: "the measurement reproduces; the conclusion does not" — both gauges are transfer-pin gauges (no residency gauge in 0.28.0); our 12 GiB tier is in fact serving heavily (counters below); 12 GiB ≈ one 150k session, not three; the native launcher still can't boot the tier — keeping open until PR #100 lands, then closing


dfein38347g:
0.28.0, single 3090, Qwen3.8-27B (GDN/mamba hybrid + MTP, 150k context, `--kv-offloading-size 24` → 12.88 GiB pinned CPU tier after the PR #100 backport boots it natively). We're re-filing per the 09-13 outcome of this thread: **the gauge reading reproduced on a clean 3090; the "tier is empty / never read" reading does not.** This version is the tracking note for the two things that *are* real — sizing guidance and the native launcher defect — and the data answers to the three asks, so we can close this the moment PR #100 lands. Thanks to @mhenrichsen for the wheel-level walk-through and @cpuchip for the three-point repro; we'll try to credit both precisely.

### 0. What this thread settled (as we read it)

- Both `kv_offload_cpu_cache_usage_perc` and the read-side gauge are **transfer-pin gauges, not residency gauges**: `num_used = allocated − free − evictable` (`kv_offload/cpu/manager.py:296–303`), and `complete_store()` (250–254) marks a block evictable the instant its copy finishes — so a resident block leaves "used" immediately. The 0.28.0 HELP string agrees: *"Fraction of CPU KV-cache space currently pinned by active transfers (0.0 = idle, 1.0 = saturated)."* mhenrichsen's 0.25-s sampling over a 48,754-token request showed **3.65 GB resident with the gauge at 0.0**, and the read gauge likewise at 0.0 right after a 1,037,303,808-byte CPU→GPU load with **27,456 external prefix hits** — which directly refutes our "never read from" reading.
- The 0.0-on-74-of-79 shape (max 2.25%) is exactly the in-flight-pin count over `num_blocks` — quanta of 1/110 and 1/444 on the two arms.
[...]

Issue #64, opened 2026-09-02
CTX=huge at MAX_LEN=240000: batched prompt_logprobs returns NaN (quality_battery 400s), plus PPL drift on the same profile


mjungnickel18:
Two symptoms on the **240k single-user profile** that turn out to share one root cause: **batched `prompt_logprobs` prefill returns NaN**, which `bench/quality_battery.py --ppl-only` surfaces as an HTTP 400 and, when it does not error outright, as run-to-run perplexity drift.

RTX 3090 (24 GB, sm_86), Windows 11 / WSL2, venv install, repo at `bb739e4`, `verify.sh --install` clean, 250 W. Service: `CTX=huge MAX_LEN=240000 PREFIX_CACHE=1`, port 18020 — i.e. our long-context profile, a thin wrapper around `single-user/start_qwen.sh`.

## 1. The error

```
{"error":{"message":"Out of range float values are not JSON compliant: nan",
          "type":"BadRequestError","param":null,"code":400}}
```

`quality_battery.py --ppl-only` failed on 2 of 4 attempts against this service. Turbo (`CTX=fast`) and the KVarN profile (`CTX=huge`, default `MAX_LEN`) never failed — 4 clean battery runs on KVarN, PPL stable to four decimals (8.4347 / 8.4349).

## 2. It is the batching, not the documents

Replaying the battery's own corpus and call shape (`/v1/completions`, `prompt_logprobs=0`, `echo=False`, `ThreadPoolExecutor(2)`), failures land on **adjacent index pairs** — which under 2 workers are exactly the co-scheduled requests:

```
run 1: 4 failures -> 16(en), 17(en), 66(da), 67(da)
run 2: 4 failures -> 16(en), 17(en), 66(da), 67(da)
run 3: 6 failures -> 16(en), 17(en), 54(da), 55(da), 66(da), 67(da)
```

Same six documents, **sent one at a time on a single thread, twice each: all six return clean logprobs.**

```
Dok 16 (en): ok / ok      Dok 54 (da): ok / ok      Dok 66 (da): ok / ok
Dok 17 (en): ok / ok      Dok 55 (da): ok / ok      Dok 67 (da): ok / ok
```

[...]

PR #100, opened 2026-09-11
Offload tier: backport the MTP/EAGLE serve fixes (#52771, #52807, #54288), boot on native from every launcher, size-checked oracle

cpuchip:
Backport of three merged upstream scheduler fixes for the CPU offload tier under MTP/EAGLE speculative decoding (vllm-project/vllm#52771, merged 2026-09-07; #52807, merged 2026-09-03; #54288, the finished-request store watermark clamp), none of which is in the 0.28.0 wheel the image pins. Plus one launcher fix without which the tier cannot boot on native Linux at all, now in all three launchers.

1. **`patches/offload-mtp-serve.patch`**: seven hunks in one file, applied after `offload-dflash-eagle-groups.patch` so the fork's DFlash narrowing stays (series line at the tail). No all-groups fallback that vetoes lookups on the Mamba layout (log once instead); the finished-request store; the lookup widened to every EAGLE group; the load boundary scanned from the computed offset; and the store watermark clamped to `num_tokens - 1`, so the tier stores the same blocks the GPU path would.
2. **`single-user/start_qwen.sh`, `batch/start_qwen.sh`, `single-user/alternative.sh`**: when `EXTRA_ARGS` carries `--kv-offloading-size` or `--kv-transfer-config`, default `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False` and print a line saying so. vLLM 0.28 refuses every KV connector under `expandable_segments:True` unless the cumem allocator is on, and the launchers set `True` on native by default, so as shipped the tier boots on WSL2 and dies on native at config validation. An explicit `PYTORCH_CUDA_ALLOC_CONF` still wins. Gotcha 38 records it.
3. **`bench/replay_offload_serve.py`**: the serve oracle behind #95 (X cold, N evictors, X again; SERVED when the second TTFT is under half the first and the CPU to GPU counter grows), with a sizing guard that refuses to report NOT-SERVED when this run's stores overflowed the tier. The guard now counts the stores once: it had summed `kv_offload_store_bytes_total` and the `GPU_to_CPU` leg of `kv_offload_total_bytes_total`, which are the same bytes, and printed INVALID-TIER-OVERFLOW on a valid SERVED run (found in review).

What the oracle says, on two layouts, both arms:

| box | route | unpatched | patched |
|---|---|---|---|
| 4090 (WSL2), MTP k=3 `CTX=long`, pool 2.5 GB, 12 GiB tier | X 22k tokens, 3 evictors | SERVED, cached 19,968, load 776 MB, TTFT 8.7 s to 1.25 s | SERVED, cached 20,800, load 806 MB, TTFT to 0.90 s |
| 3090 (native), MTP `CTX=long`, `KV_MEM` 1 GiB, 16 GiB tier, `PREFIX_CACHE=1` | X 125k chars, 3 evictors, ~120k tokens of distinct prefill | SERVED, ratio 0.08, loads 1.07 GB, external hits 28,288 | SERVED, ratio 0.05, loads 1.10 GB, external hits 29,120 |

The watermark clamp has since been measured on both boxes against the 0.29 series (issue #106): the cached count on the replay grows by exactly one 832-token block on each (WSL2 16,640 to 17,472; native 11,648 to 12,480), which is the block formula the GPU path uses, `(floor(n/B) - 1) * B`. That is the "tier path one block short" shape from the #95 thread, closed.

So the upstream veto does not reproduce on this model's layout on either box, and the backport is no-harm here beyond the one block. It ships because the wheel is behind three merged fixes that target exactly this configuration, and because the launcher fix is needed regardless.

Sizing, with the maintainer's numbers from #95 rather than mine: 12 GiB holds about one 150k session on `CTX=long` (444 blocks of ~336 tokens; 86.2 KB per tier token against 37.9 KB per GPU-pool token on that box, a 2.3x ratio), so at the launcher's default pool on a 3090 the 12 GiB tier is smaller than the GPU prefix cache and cannot hold what the GPU cache has already dropped. The tier's `kv_offload_cpu_cache_usage_perc` gauge counts in-flight transfer pins, not resident blocks, so a 0% reading is not an empty tier; residency has no gauge in 0.28, read `kv_offload_total_bytes_total` and the external-prefix-cache hit counter instead.

Verified on the native 3090 against this head (one card at 250 W, production stopped): `batch/start_qwen.sh` and `single-user/alternative.sh` each booted with `EXTRA_ARGS='--kv-offloading-size 8'` (alternative.sh with `MAX_LEN=40000`; its default 256k does not fit beside an 8 GiB tier on 24 GB and refuses at KV sizing, which is capacity, not the allocator), the notice line printed exactly once, `OffloadingConnector` and `CPUOffloadingSpec` created, and `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:False` read from the engine's own `/proc` environment rather than the launcher's echo; the control, the same launcher and environment from main 8d09ec5, died in about 20 s with the `expandable_segments:True unless enable_cumem_allocator` ValidationError and no notice line. The oracle with the corrected guard on the mtp/long tier profile: SERVED, stores 7.88 GB against an 8.0 GiB tier with no overflow verdict, cold 24.65 s to replay 2.72 s, cached 22,464 (27 blocks of 832); at that margin the old double count would have read about 15.8 GB and raised a false overflow, so the fix is shown by the margin rather than asserted. Series integrity at this head against a pristine v0.28.0: the patch applies in its series position with 0 offsets and 0 fuzz; 33 patches applied, 0 rejected.

What is still unproven: the native control was run for `alternative.sh` and not for `batch/start_qwen.sh` (same block, so inferred rather than measured); no serve run from those two launchers, only the boot; `single-user/start_qwen.sh` with the tier was proven in the 0.29 work, not in this pass; every prefix hit on the maintainer's CTX=huge run arrived through the tier, so the new store assert was only ever evaluated at a block-aligned local count of 0 (the caller-side LCM alignment is the reason it cannot fire elsewhere, not a measurement); no quality battery on the offload profile (CTX=huge with MTP corrupts `prompt_logprobs`, #64); the launcher fix in `batch/` and `alternative.sh` is verified by boot on native, not by a serve run from those two launchers; the 16 GiB tier used on the 3090 row does not boot on a 30 GB host (`/dev/shm` at 15.6 GB), so the native rows are at 8 GiB.


cpuchip:
One more upstream fix that belongs with this: vllm-project/vllm#54288 (merged 2026-09-06, not in 0.29.0 either). A finished request offloads up to `num_tokens`, and the final sampled token's slot was never written by any forward pass, so the tier stores a block whose last slot is junk under a hash of legitimate token ids; a later turn that replays those tokens loads it, and under speculative decoding the model attends to a token that was never accepted. Same scheduler file as the two hunks here. Reproducing it on this model first, then folding it in as a third hunk on this branch.

cpuchip:
Folded in as the seventh hunk (3284992). What verified it, on two boxes: upstream's offloading scheduler tests at v0.28.0 with the test changes of the three PRs this branch backports layered in. As shipped, before this hunk, 115 of 121 pass; the six failures are the two EAGLE trailing-block tests and the four tests #54288 rewrote. With the hunk, 121 of 121, no regressions. Identical on a native 3090 venv (0.28.0 plus the fork series) and on the fork image on a WSL2 4090. The trailing-block failures are the reason it belongs here: upstream wrote #52771 against a tree that already carried #54288, so the backport without it fails upstream's own contract.

On the real path it is reachable only with `offload_prompt_only: false` in a full connector config (the `--kv-offloading-size` shortcut clamps stores to the prompt, and every prompt token is written by prefill). Even then, on this hybrid layout a follow-up request's prefix hit never reaches the last several hundred to sixteen hundred tokens of the previous request (one Mamba state snapshot per turn at the last chunk-step boundary, and a hit floors to it), so the final block is never replayed and there is nothing to observe here. Measured with a token-id replay through `/v1/completions`: the tier stored and served every replay, cached tokens stopped inside that tail every time, outputs and logprobs identical with and without the hunk. It is carried for correctness, not for a symptom anyone on this model can hit today.

mhenrichsen:
Holding this open for three small fixes, not a rework. The assert question — the one thing I actually held it for — is settled in your favour, on the box.

**The assert is fine, and here is why.** I asked you to justify `assert num_locally_computed_tokens % tokens_per_block == 0` against this fork's geometry, where `--prefix-match-unit 128` (CTX=huge) is deliberately smaller than a group block and the kvarn coordinator assert was relaxed to allow it. The caller already guarantees it. Installed vLLM 0.28.0, `vllm/v1/core/sched/scheduler.py` ~820:

```
partial_tail = num_new_local_computed_tokens % self.block_size
block_aligned_local = num_new_local_computed_tokens - partial_tail
ext_tokens, load_kv_async = self.connector.get_num_new_matched_tokens(request, block_aligned_local)
```

`self.block_size` is the scheduler block size (LCM of the group block sizes), so what reaches the connector is aligned to every group's `tokens_per_block`. That line is upstream, not ours — `grep -n "Present a block-aligned local hit" patches/*.patch kvarn/*.patch` is empty. The 128-granular prefix-match-unit feeds `block_hashes`, not this number.

**GPU arm**, one 3090, production stopped, scheduler patched, CTX=huge SPEC=mtp PREFIX_CACHE=1 MAX_LEN=24576 MAX_SEQS=2, `EXTRA_ARGS='--kv-offloading-size 8 --kv-cache-memory 2000000000'`. Geometry confirmed live in the engine args: `block_size 128`, `prefix_match_unit 128`, `kvarn_k4v2_g128`, and `interface.py:928 Setting attention block size to 2048 tokens` — hash unit 128, group block 2048, i.e. exactly the sub-block layout in question. 14 requests including a token-exact 128-aligned / non-2048-aligned prefix (12,160 templated tokens, `12160 % 2048 = 1920`, served hit 12,032 = 94x128):

```
{"req":"R1_PQ","s":19.76,"prompt_tokens":20287,"cached":0,"ext_hit":0.0}
{"req":"R2_P_serve","s":1.87,"prompt_tokens":12066,"cached":10240,"ext_hit":10240.0}
{"req":"R2b_P_128aligned","s":0.26,"prompt_tokens":12160,"cached":12032,"ext_hit":93696.0}
```

`grep -c AssertionError` -> 0; no `EngineCore died`/`Traceback`; `/health` 200 after all 14. And the tier genuinely serves: 19.76 s cold prefill vs 1.87 s after eviction, `vllm:kv_offload_total_bytes_total{transfer_type="CPU_to_GPU"}` 0 -> 8.29e+08. Not the "stores but never serves" shape.

**The launcher hunk reproduces exactly as you described it.** main-branch launcher, CTX=long SPEC=mtp, `EXTRA_ARGS='--kv-offloading-size 16'`:

```
ValidationError ... KV connector OffloadingConnector is incompatible with
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True unless enable_cumem_allocator is also enabled
EXIT=1
```

No server at all on native Linux. With your hunk it boots and prints the notice; with `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` set by hand the notice is suppressed (`grep -c` -> 0) and the boot fails as before — so it is a default, not a clamp. Hunk 1 also fires and `logger.info_once` is not an AttributeError: `scheduler.py:251 KV offloading: speculative decoding is enabled but no KV-cache group is annotated as a drafter group`.

**What is still unproven**, stated plainly so nobody reads more into the above than is there:
- Every prefix hit on this configuration arrived through the tier, not the GPU (`vllm:external_prefix_cache_queries_total` grew by the full prompt length on every request, including an immediate repeat), so `block_aligned_local` was 0 throughout and the new assert was only ever evaluated at 0. The reason to believe it cannot fire at a nonzero value is the caller-side LCM alignment above, not the 14 requests.
- No quality check. `bench/quality_battery.py` was not run, and CTX=huge + SPEC=mtp + PREFIX_CACHE=1 corrupts `prompt_logprobs` anyway (#64), so that has to be measured on a different profile.
- I did not re-confirm your own (honestly self-reported) claim that the upstream veto does not reproduce on this layout without the patch — I only confirmed the tier serves with it.
- Tier 8 GiB, not 16: `--kv-offloading-size 16` cannot boot here, `Insufficient space in /dev/shm: 16372 MiB required, 15620 MiB free`. With a pinned 2 GB pool that is a small-pool high-churn regime; the absolute latencies are not production numbers.

**To unblock the merge**, three things, none of them touching the patch:
1. Refresh the PR body and the lead commit message — both still say "two merged upstream fixes" / "the four hunks from upstream"; since 3284992 it is three fixes and seven hunks. The patch header is already correct.
2. Port the ALLOC_DEFAULT hunk to `batch/start_qwen.sh` (lines 159-165) and `single-user/alternative.sh` (lines 14-19). Both carry the same block and both expand `EXTRA_ARGS`, so the tier still cannot boot on native from either.
3. Update gotcha 38 in `docs/gotchas.md`. It currently tells the reader to set `expandable_segments:False` by hand and says only the WSL2 branch defaults to it — that now contradicts the branch. Record that the launcher does it automatically for a KV connector, and that the MTP serve backport shipped.

Two notes for elsewhere, not for this PR. The `kv_offload_cpu_cache_usage_perc` gauge counts in-flight transfer pins, not resident blocks, so the "0.0 on 74 of 79 samples" in #95 is not evidence of an empty tier — that explanation is what closes #95, not this patch. And `docs/MR-DRAFT.md`'s owed-checks list (section 3) belongs to the `VLLM_INT4_MQ_3D` default flip; this PR neither touches nor discharges any of it.

Credit where it is due: the diff does what it says, all seven hunks are on the head commit, patch-integrity is green, the alphabetical apply order against `offload-dflash-eagle-groups.patch` holds, and you wrote up front that the upstream veto does not reproduce on this layout and that the #54288 hunk is unobservable here. That last part is rarer than it should be and it is the reason this took one GPU run instead of three.

cpuchip:
Pushed, all three asks and the two notes:

1. Body and lead commit now say three fixes and seven hunks (#52771, #52807, #54288); the branch is rebased onto 8d09ec5 and `patches/series` lists `offload-mtp-serve.patch` at the tail (after `offload-dflash-eagle-groups.patch`, line 32).
2. The allocator default is in `batch/start_qwen.sh` and `single-user/alternative.sh` as well. Verified on a native 3090 against this head: both boot with `--kv-offloading-size 8` in `EXTRA_ARGS` (alternative.sh with `MAX_LEN=40000`, since its 256k default does not fit beside an 8 GiB tier on 24 GB), the notice prints once, and `expandable_segments:False` is read from the engine's own environment; the same launchers from main die at config validation with the ValidationError you quoted.
3. Gotcha 38 says the launchers default it and that the MTP serve backport is in the series.

Your note on the oracle guard was right: `store_gb()` summed the store counter and the GPU_to_CPU transfer counter, the same bytes twice. It counts once now. On the native mtp/long tier profile the corrected oracle reads SERVED with stores at 7.88 GB against an 8.0 GiB tier, a margin at which the old double count would have raised the false overflow; replay 24.65 s to 2.72 s, cached 22,464 (27 blocks of 832).

Still unproven on this pass: the native control was run for alternative.sh and inferred for batch (same block); the two sibling launchers were booted, not served from; `start_qwen.sh` with the tier was verified in the 0.29 work rather than here. Your #95 sizing numbers are in the body in place of my 100k.

mhenrichsen:
Merging this.

Round two delivered all three of the things I held it for, and I checked them on the head (6a9b8c1) in a fresh clone rather than from the thread:

- Scope matches the text. `grep -c '^@@' patches/offload-mtp-serve.patch` = 7, and `grep '^+++'` shows one target file, `distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py`. Body and lead commit now both say three fixes / seven hunks.
- The allocator default is in all three launchers. `grep -rln 'KV connector in EXTRA_ARGS'` returns `single-user/start_qwen.sh`, `single-user/alternative.sh`, `batch/start_qwen.sh` — exactly the set that defines `ALLOC_DEFAULT`.
- Gotcha 38 is no longer stale: it says all three launchers default it, that an explicit `PYTORCH_CUDA_ALLOC_CONF` still wins (so setting `True` by hand reproduces the refusal), and that `offload-mtp-serve.patch` is in the series.

Series and CI: `patches/series` lists `offload-mtp-serve.patch` at line 48, after `offload-dflash-eagle-groups.patch` at line 32, which is the order the patch header requires. The patch-integrity job runs `bash patches/check_vllm_series.sh` against a pristine `vllm-project/vllm@v0.28.0` checkout and is green on this head, so series-vs-directory equality and every hunk are checked against real upstream source, not a subdirectory.

The one thing I went looking for on my own was a `set -e` trap. `single-user/alternative.sh` sets `set -e` at line 11 and your new block is an `[ ] && [ ] && echo` AND-list — this repo has been bitten by exactly that shape before (#59, whose warning comment sits in all three of these files). It is safe. Reproduced all three branches standalone (no override, explicit `expandable_segments:True`, WSL2 default already `False`): every one exits 0 and reaches the end of the script, because the AND-list is not the last command in the `case` branch. Good instinct putting the assignment after the notice rather than folding it into the chain.

No GPU claim outstanding. The correctness claim was reproduced-and-absent on the box last round — ValidationError with main's launcher, boot plus the notice with the hunk, notice suppressed and boot failing again with `expandable_segments:True` set by hand, which is what makes this a default and not a clamp — and the two sibling launchers booted native against this head with `--kv-offloading-size 8`. There is no throughput claim to A/B and no default flip in the product path: all seven hunks live inside the offloading connector's scheduler, dead code unless a KV connector is configured, and production is not. That is also why I am not asking for a quality battery.

Still unproven, and I am landing it anyway because none of it gates a code path production instantiates: no actual serve run from `batch/` or `alternative.sh`, only boot; the new store assert has only ever been evaluated at a block-aligned local count of 0, justified by the caller-side LCM alignment rather than measured; and no quality numbers on the offload profile, since #64 poisons `prompt_logprobs` on CTX=huge+MTP so there is no trustworthy instrument there yet. Your residual list said all of that, which is the reason this is a two-round review and not a five-round one.

One cosmetic I am fixing at merge instead of sending back for a force-push: the lead commit body (9c01d39) still reads "(both land after the 0.28.0 wheel)" in the sentence that now enumerates three fixes.

Separately, not on this PR: the `kv_offload_cpu_cache_usage_perc` gauge counting in-flight transfer pins is what actually closes #95. Your corrected `store_gb()` — it was double-counting `kv_offload_store_bytes_total` against the GPU_to_CPU leg of `kv_offload_total_bytes_total` — is already in this diff, so that is one less thing in the way.

Issue #106, opened 2026-09-13
vLLM 0.29.0 port is ready on our fork (patch series, KVarN, #100 and #101 carried); PR when you want it


cpuchip:
@mhenrichsen, the 0.29.0 port is done and vetted on two boxes, and I would rather not drop a PR that touches every patch file on you unannounced. It sits on our fork as `port-0.29`, cut from your main at 1834917 so it carries only the pin:

- branch: https://github.com/cpuchip/qwen38-27b-rtx3090/tree/port-0.29 (head `8544dbf`, 24 commits over 1834917)
- the ledger: `PATCHES.md`, one line per patch (kind, what, upstream reference, what retires it)
- the port notes: `docs/vllm-0.29.md`

Say when you want the PR and I will open it against main; if you would rather pull the branch directly, it is yours.

## What it carries

1. `vllm==0.29.0` and the dependency pins 0.29 requires. Two patches retired because 0.29 carries them (#54282 Gumbel-salt draft decoupling, the xgrammar terminated fix); two hunks retired for the same reason (the graph-memory reserve, the int4 padded-page view); two regenerated (dflash2 prewarm, lookup drafting); two adjusted for 0.29 API changes (`dp_sync` on the ngram-chains `propose` override; the block-promotion pad check now mirrors upstream's own).
2. KVarN on the 0.29 layout contract: the backend declares its layout (LBHNC) and folds the runner's per-layer view into one tile per block and head; a wrong layout now fails at the first KV update instead of returning wrong numbers.
3. #100 and #101, carried as 0.29 patches, because the 0.29.0 tag has neither upstream fix: `offload-mtp-serve` (#52771/#52807, all seven hunks unchanged) and `mamba-align-retire-null-gaps` (#55450, two hunks re-anchored). This branch supersedes both PRs; I can close them when this lands.
4. Every patch file is regenerated against the 0.29.0 tree and applies with exact context: 30 clean, 1 at an offset, 0 with fuzz. The Dockerfile and `check_vllm_series.sh` now apply with `patch --fuzz 0`, so a hunk whose context has moved fails the build by name instead of landing by guess. (On the 0.28 series applied to 0.29, nine patches were landing by fuzz and nothing said so.) The CI workflow reads the pin from `docker/requirements.txt` instead of a hardcoded tag.
5. The launcher defaults `expandable_segments` off when a KV connector is configured (from #100; without it the offload profile cannot boot on a native box).

## How it was vetted

[...]

PR #108, opened 2026-09-13
gotchas: the large-pool limit beside item 12, the null-gap retirement beside item 44

cpuchip:
Two documentation follow-ups asked for in the #91 and #101 merge comments, nothing else: the large-pool limit beside gotcha 12 (the int32 cast, the ~2,383-block threshold, the synthetic test, the same pattern in the int8 prefill kernel) and the null-gap retirement beside gotcha 44 (your 16 to 5 retained-block counter and your four-decimal CTX=fast peaks from the #101 review, attributed; the 30 to 50 percent range across depths and pins from a second native 3090 and a WSL2 box with the pool pinned across arms; the single-space blank context lines a whitespace trim destroys). No code, no patch, no launcher. Numbers are yours from those two comments plus the two-box range already reported on #106. What is still unproven: nothing new is measured here.


mhenrichsen:
Merging, with one clause edited at merge time (70faf54, pushed to your branch).

Everything else here checks out against the sources, not against the PR body:

- Item 44: `patches/mamba-align-retire-null-gaps.patch` is on main and is line 47 of `patches/series`. The "16 before, 5 on every request" counter, the -40.5% no-cached-prefix row, the -21.1/-22.8% cached rows and the four-decimals-identical CTX=fast peaks over 20 requests are my own #101 merge comment verbatim. The "30 to 50 percent" two-box range spans #101 (36.8->21.2, 68.9->40.6, 32.2->19.1, 44.8->23.9, 60.4->30.4) and #106's 0.29 table (35.2->21.0, 67.6->40.5, 24.3->16.8, 48.0->30.2), so both the range and the "0.28 and 0.29" attribution are sourced.
- The single-space blank context lines are real and worth documenting: `grep -c '^ $' patches/mamba-align-retire-null-gaps.patch` -> 2, which is exactly what 07f791a had to restore.
- Item 12: the int32->int64 story, the ~2,383-block threshold (bf16 stride 880*4*256 = 901,120 elements), the 64 GB-only reach and the test all hold. `grep -n 'tl.int64' patches/spec-decode-attn.patch` -> line 225, `bench/test_spec_decode_bigpool.py` is present.

The one thing that was not true on main: "the same pattern in the int8 prefill kernel has the same fix and its own test". `grep -c tl.int64 patches/prefill-attn-int8.patch` -> 0, and `bench/test_prefill_attn_bigpool.py` does not exist. That cast and that test are #109, which you opened 47 minutes after this branch and which is still open. gotchas.md is the one file whose whole job is to be honest about what main does, so I changed the clause to "The int8 prefill kernel carries the same pattern; its cast and test are pending in #109." I will drop the qualifier when #109 lands; #109 gets its own triage rather than riding in behind a doc PR.

Non-blocking, for whoever edits that paragraph next: it says the #101 pool was "pinned identical across arms". On the reference 3090 I verified the pool came out identical, which is not the same as pinning it, and #106 records peak-pool numbers on that box moving 59% between sessions. Keep the within-session caveat if anyone tightens that sentence.

CI green, no rebase needed, no series entry needed. Thanks - this is exactly what the #91 and #101 merge comments asked for, and the numbers are attributed correctly.

Issue #94, opened 2026-09-10
Requests can leave the running set with no completion, no abort, and no error line (0.28.0 logs zero completion lines): 15-min client holds, a 28-min held-not-lost wedge caught mid-stall — single 3090, Qwen3.8-27B (the request-side half of #107)


dfein38347g:
0.28.0, single 3090, Qwen3.8-27B (GDN/mamba hybrid + MTP, 150k context). This is the **request-side half** of the stall work tracked in #107: what a *client* experiences when a request stops being served — and the logging gap that makes the whole class invisible. We're re-filing after the 09-11→09-13 discussion in this thread, which changed the shape of what we can and can't claim. Thanks to @cpuchip and @mhenrichsen for the path inventory, the abort-line flag discovery, and the reference-3090 deaths — the rewrite below stands on that work, and we'll try to credit it precisely. We're probably still missing context you'd have — we're just users on one 3090 asking for help — and any pointer saves us days.

### 0. Status of the original event (09-10 16:59:00→17:00:00Z) — **unclassified**, per the thread

The original filing (a 150k request that died mid-stream: KV cache freed down to `Running:0` inside one 10-s stats interval; router `recovered from upstream disconnection` with **no** `httputil: ReverseProxy read error` line in that window — per cpuchip's 09-11 taxonomy that points at the client-close / no-upstream-error form, not an upstream read error; the client saw a `Connection error` ~4 min into the stream; raw window: [gist 95091970…](https://gist.github.com/dfein38347g/95091970f332288983e46f2d6fbe7160)). Per mhenrichsen's 09-13 decision and cpuchip's 22:23Z self-correction: **0.28.0 log silence does not discriminate between a client-side abort and an engine death** (three silent EngineCore deaths on the reference 3090, below, are the direct evidence), so the event stands **unclassified** until the total-request budget or the two abort lines say otherwise. The 09-13 15:26:57Z item we'd offered as a "repro" was retracted in-thread (it completed 200 in 11.9 s) and stays in the record.

[...]

PR #110, opened 2026-09-15
patches: engine completion line and stall sentinel (#94, #107)

mhenrichsen:
Carries the ask-3 draft from #107 (gist `9f1511885`, two diffs against the v0.28.0 tag) into `patches/`, with three fixes, and reports one boot of measurement on the reference 3090.

## Why

0.28.0 has no request-completion log line at all — not gated, absent. The only request-lifecycle lines in `v1/` are `Added request`, `Aborted request(s)` and `Request ... aborted.`, and all three sit under `if self.log_requests:` (`v1/engine/async_llm.py:438, 760, 622`), i.e. behind `--enable-log-requests`, off by default. That is the 1473 Added / 0 finished / 25 Aborted in 21.5 h reported on #94, and it is why the 09-10 event there is unclassifiable after the fact.

## Fixes to the draft

1. **The sentinel as drafted does not boot.** It is constructed beside `identity = self.engine_index.to_bytes(...)` in `EngineCoreProc.__init__`, and reads `self.scheduler.get_request_counts` eagerly — but `self.scheduler` is created by `super().__init__()`, ~115 lines later. Observed on the box, 2026-09-15 08:32:21Z:
   ```
   (EngineCore pid=29244) ERROR 09-15 08:32:21 [core.py:1423] AttributeError: 'EngineCoreProc' object has no attribute 'scheduler'
   (APIServer pid=29152) RuntimeError: Engine core initialization failed.
   ```
   Moved to directly after `super().__init__()`, which also means the thread starts after model load rather than before it.
2. **The knob is registered in `vllm/envs.py`** (annotation + `environment_variables` lambda) instead of `os.getenv`, per @cpuchip's note on #107.
3. **The sentinel calls `get_request_counts()` itself each tick** rather than reading a snapshot the main loop refreshes, so a core that stops stepping with a stale non-zero snapshot is reported against its current state. This deletes the `step_started` hook entirely — the patch now touches one call site instead of two.

Also dropped the draft's `+import threading`: 0.28.0's `core.py` already imports it at line 7, and the draft adds a second one in the `from multiprocessing.queues import Queue` block, which fails ruff/isort.

## Measurement

One boot, `CTX=fast SPEC=dflash2 DFLASH_TOKENS=15`, `--enable-log-requests`, `VLLM_ENGINE_STALL_SENTINEL_S=30`. All three arms in that one boot.

**Arm A — 20 sequential streaming completions, `max_tokens=400`.** 20 lines, all matching the client:
```
INFO 09-15 08:39:00 [scheduler.py:1984] Request finished: req chatcmpl-b4e8ef66c5bdd0b9-... finish_reason=length prompt_tokens=24 generated_tokens=400 elapsed_s=1.36
... (20 total; every generated_tokens=400 == the client's usage.completion_tokens,
    every elapsed_s within 0.02 s of client wall time)
```

**Arm B — one streaming request, client SIGKILLed at 5.01 s after 192 chunks.** The discriminator, demonstrated rather than argued:
```
INFO 09-15 08:39:25 [async_llm.py:439] Added request chatcmpl-a82973ac6bc86577-92c4d50d.
INFO 09-15 08:39:30 [async_llm.py:761] Aborted request(s) chatcmpl-a82973ac6bc86577-92c4d50d.
INFO 09-15 08:39:30 [async_llm.py:623] Request chatcmpl-a82973ac6bc86577 aborted.
```
`grep -c "Request finished: req chatcmpl-a82973ac6bc86577"` → **0**. A client death and a completion are now distinguishable in one read.

**Arm C — one 44,820-token prompt, `max_tokens=16`, 28.5 s of chunked prefill.**
```
INFO 09-15 08:40:02 [scheduler.py:1984] Request finished: req chatcmpl-81ff4418e524e220-99ce6b38 finish_reason=stop prompt_tokens=44820 generated_tokens=3 elapsed_s=28.48
```

**Totals for the boot:** 22 `Added` / **21** `Request finished` (20 A + 1 C; arm B correctly absent) / 2 abort lines / **0** `EngineCore stall:` / 0 tracebacks, `NameError`s or `ERROR` lines.

The zero stall lines on arm C are the point: this sentinel keys on *completed steps*, not on stats-line silence, so a chunked prefill steps per chunk and never trips the clock however long it runs. That is the false-positive class that gave the external log-scraper 80+ hits, and it does not exist here.

## Caveat

No wedge was induced, so the sentinel's **true-positive** path is unexercised on this box — only its quiet-under-load behaviour and its boot are measured. The completion line is exercised end to end.

The patches are validated to apply with `patch -p1` as the last two lines of `patches/series`, and the resulting tree is byte-identical to the one the numbers above were measured on.


dfein38347g:
A note on the sentinel knob, from the side that proposed it. As merged, `VLLM_ENGINE_STALL_SENTINEL_S` is registered in `vllm/envs.py` without an `ignored_factors` entry, so — per your point (v) — any threshold change lands in the torch.compile cache key and costs a full cold recompile (~2.7 min on your box) before it can be used. It's a diagnostic-only knob, off by default, and it gates no model code, so we'd like it excluded from the compile cache key: the original draft here registered it as an ignored factor, and the merged version appears to have dropped that. If it's a two-line change, happy to send the patch on this issue; if you'd rather keep the registration plain, no hard feelings — the tax only bites whoever actually changes the threshold mid-session.

PR #116, opened 2026-09-15
Harden chat templates against array tool-call arguments

mhenrichsen:
The template half of #85, split out for its own review. Both commits are Anton Rastyazhenko's work, authorship preserved:

- `bd716f27` — `prepare/harden_chat_template.py` plus the `docker/prepare.sh` call. Some clients (JetBrains AI Assistant) send assistant `tool_calls` with `function.arguments` as a JSON *array*; the Qwen3 template renders arguments with `tool_call.arguments|items` and jinja2 raises `Can only get item pairs from a mapping.`, so an otherwise valid request 400s. The rewrite branches on `arguments is mapping` and falls back to a single positional parameter. Real bug, right shape of fix.
- the `HARDEN_TEMPLATES=0` opt-out, cherry-picked from `c063cd12`. The `NO_DOWNLOAD` / `BASE` fallback hunks of that commit are unrelated scope and are not taken here.

One thing I want changed before this merges, and it is why the split was worth doing:

**An unrecognised template should warn, not fail.** `main()` sets `rc = 1` whenever `res == "unknown"`, the call sits before the final `state`/`LEFT` check, and `docker/prepare.sh` runs under `set -e`. The default is still `HARDEN_TEMPLATES=1`, so the day upstream revises the Qwen3 template, `prepare` exits non-zero *after* the ~19.5 GB download and the requantisation have already succeeded, and the user is left with a failed prepare and a ready model dir. Keep `rc = 1` for the `OSError` branch — a template that cannot be read is a real failure — and make "unknown" a warning. The `HARDEN_TEMPLATES=0` opt-out does not cover this: nobody sets it before the failure they have not had yet.

Retracting one thing I said on #85: the `python`-from-`PATH` point does **not** apply to the Docker path, `docker/prepare.sh:11` exports `/app/venv/bin` first. It only holds for the bare-clone native README path.

Not carried over from #85 and not mine to decide: the `docker-compose.yml` bind-mount of `./docker/prepare.sh` over `/app/docker/prepare.sh:ro`. It pins every compose user's prepare step to their checkout rather than the image they built, which is a real behaviour change and wants its own justification. Open it separately if you want it.


mhenrichsen:
Made the change I asked for myself rather than leaving the PR sitting on it — `fa73150`.

`"unknown"` now warns on stderr and leaves the template alone; `rc = 1` stays for the `OSError` branch, where a template that cannot be read is a real failure. The `docker/prepare.sh` comment said the opposite and is corrected with it.

Checked all four cells locally (no GPU needed, it is on a training job):

- unrecognised template → `warning: unknown: … — left as-is; array tool-call arguments may still 400`, **rc=0**
- unreadable template (mode 000) → `[Errno 13] Permission denied`, **rc=1**
- known template → `hardened`, rc=0
- same template again → `unchanged`, rc=0 (idempotent)

Merging. Both original commits keep Anton Rastyazhenko's authorship.

Still not carried over, and still not mine to decide: the `docker-compose.yml` bind-mount of `./docker/prepare.sh` over `/app/docker/prepare.sh:ro`. Open it separately if you want it.

Issue #114, opened 2026-09-15
spec-decode-attn.patch: VLLM_SPEC_DECODE_ATTN_QMAX and VLLM_SPEC_ATTN_BLOCK_M are read with bare os.environ.get, outside the compile cache key


mhenrichsen:
#90 fixed `VLLM_SPEC_ATTN_DEBUG`: it was read with a bare `os.environ.get`, so it sat outside vLLM's env registry and outside the torch.compile cache key, and the string `"0"` is truthy — `VLLM_SPEC_ATTN_DEBUG=0` turned the debug path on. Two knobs in `patches/spec-decode-attn.patch` are still in that state:

- `patches/spec-decode-attn.patch:109` — `n = int(os.environ.get("VLLM_SPEC_DECODE_ATTN_QMAX", n or 0)) or n`
- `patches/spec-decode-attn.patch:298` — `override = os.environ.get("VLLM_SPEC_ATTN_BLOCK_M")`

(`:85`, `VLLM_SPEC_DECODE_ATTN`, is the same class but compares `== "1"`, so at least its truthiness is right; it is still unregistered.)

Consequence is worse than the one just fixed. `VLLM_SPEC_ATTN_BLOCK_M` changes the kernel launch shape and `VLLM_SPEC_DECODE_ATTN_QMAX` changes the query-block cap, and neither is in the compile cache key — so two runs that differ only by one of these knobs can hit the same cached artifact. We have been bitten by an env that is not in the cache key before (the envs-in-cache-key fix on the DFlash3 box).

Fix: register all three in the `envs.py` hunk of `patches/spec-decode-attn.patch` the way `speed-knobs-envs.patch` and now the fp8 patch do, and read them through `vllm.envs` at the use sites. Ints go through the registry as ints (`VLLM_SPEC_DECODE_ATTN_QMAX`, `VLLM_SPEC_ATTN_BLOCK_M` as optional int), not as strings.

Verification that would close this: with the patched tree, the same request and profile with only the knob differing produces a different compiled artifact (or at minimum a different cache key), and `=0` / `""` / unset all read False for the bool; then `bash patches/check_vllm_series.sh <pristine vLLM v0.28.0 checkout>` green.

cpuchip:
[...]

PR #119, opened 2026-09-15
Series hygiene on 0.28: patches regenerated from fork commits, fuzz-0 apply and check, CI on the pinned tag, PATCHES.md, the knobs each patch reads registered (#106 part one, #114)

cpuchip:
The pin-independent half of #106, against main on 0.28. Six commits (the fourth is one line: the series check's usage string names the package directory, since the checkout root reads as a series break; the fifth adds `--no-backup-if-mismatch` to the Dockerfile's apply loop, since a hunk applied at an offset otherwise leaves a `.orig` in site-packages and the series check already passes the flag; the sixth is a comment wording change in `verify.sh`).

1. **Every patch file regenerated from a fork commit, and the tooling applies and checks at `--fuzz 0`.** On a pristine v0.28.0 tree, six of main's patches applied only with fuzz (hybrid-sw-block-promote, vision-tower-cpu-offload, offload-dflash-eagle-groups, offload-wsl2-devptr, spec-sampler-prewarm, vllm-pr54282-draft-gumbel-salt) and 21 more at an offset; `check_vllm_series.sh` counted both under one word and failed on neither. The files are now exported from `cpuchip/vllm` `qwen38/0.28` (v0.28.0 + one commit per patch, series order, subject `[qwen38] <topic>`), so each applies with exact context at its position. The Dockerfile's apply loop, the series check, `kvarn/install.sh` and `verify.sh` run at `--fuzz 0`; the installer stops on a FAILED hunk instead of `|| true`; `verify.sh` asks the live registry for the int8 knob instead of grepping the file.
2. **CI checks the series against the pin in `docker/requirements.txt`**, not a hard-coded tag, so a pin bump and its regenerated patches are judged together.
3. **`PATCHES.md`**, one line per patch (kind, what, upstream reference, cut against, what retires it), and `scripts/export-patch.sh`, which regenerates a file from its fork commit.

**The knob registration (#114 and the rest of that family).** By set difference against pristine v0.28.0, main's series reads 24 `VLLM_` names it never registers and 5 it registers but reads raw (`os.environ` beside the registry). Each patch now registers the knobs it reads in its own `envs.py` hunk and reads them through `vllm.envs`, which puts them in the torch.compile cache key. `VLLM_SPEC_DECODE_ATTN` and `VLLM_DRAFT_TOPK_TOPP` move out of `speed-knobs-envs` into `spec-decode-attn` and the sampler patch, which read them; `speed-knobs-envs` keeps the Marlin tune pair. Defaults were checked cell for cell (29 knobs, 7 environment values each: unset, empty, 0, 1, 16, int8, abc); two cells differ and both are intended: `VLLM_SPEC_ATTN_BLOCK_M=0` now means "pick by row count" instead of forcing a zero-row tile, and an empty `VLLM_SPEC_DECODE_ATTN_QMAX` falls back to the drafter's block instead of raising. `KVARN_*` names are not in this pass: `kvarn/files` are copied modules, not patches, and the unknown-variable warning fires for `VLLM_` names only; the 0.29 port registers them.

**Checked, and how.**

- The exported set replayed at `--fuzz 0` onto pristine v0.28.0, series order then `kvarn/install.sh`'s steps, reproduces the fork branch tree byte for byte (0 differing files).
- Against main's files: 27 patches keep the same added and removed lines exactly; 13 differ only by the sweep (the raw read replaced by the registry read, the `envs.py` registration lines, the `envs` import, the two moved knobs, and the fp8 patch no longer adding an import that `spec-decode-attn` now owns). One apparent difference is main's `spec-decode-scratch-within-budget` removing and re-adding an identical line, which git folds into context.
- `patches/check_vllm_series.sh` on pristine v0.28.0: pass 1, 38 patches applied with exact context, 0 at an offset, 0 with fuzz; pass 2, the five contractual DFlash patches apply standalone with `git apply --check --whitespace=error` (the two that register knobs anchor on a region of `envs.py` no other patch touches, so the standalone contract holds).
- The image built from this branch (`--fuzz 0` apply loop, `kvarn/install.sh`, `verify.sh --install`: OK, 0 failures) has a `vllm` .py tree identical to the fork branch tip, wheel-generated files excepted.
- Boot on WSL2 (RTX 4090), the production line `SPEC=dflash2 CTX=fast PREFIX_CACHE=1 DFLASH_TOKENS=15 INT8_ACT=int8 PREFILL_ATTN=int8`, three boots in one session: main bae2023 as-is, health 200, `GPU KV cache size: 57,669 tokens`, 4 unknown-variable lines (`VLLM_V2_CUDAGRAPH_MEM_MIB`, `VLLM_SPEC_DECODE_ATTN_QMAX`, `VLLM_PREFILL_ATTN`, `VLLM_DFLASH2_LOOKUP`); this branch, health 200, 57,669 tokens, 0 unknown-variable lines; this branch with one made-up `VLLM_` name exported as the positive control, health 200, 57,669 tokens, exactly 1 line naming it. Same block size (480) on all three.
- Native 3090 (Linux, python 3.14.4 on that host, the same 0.28.0 that serves production there): `patches/check_vllm_series.sh` both passes, exit 0 (pass 2 is the one a Windows path cannot run). Fresh tree from the branch, own venv from `docker/requirements.txt`, 38 patches at `--fuzz 0`, kvarn installed, `verify: OK`; the production line boots to health 200 with 0 unknown-variable lines, and the same boot with one made-up `VLLM_` name exported prints exactly 1 line naming it. Read separately from the tree rather than from this text: 35 registry keys on the tip that v0.28.0 lacks, of which 29 are this PR's (24 newly registered, 5 whose raw read moved to the registry) and 6 were already on main (the int4 pair, `VLLM_MARLIN_TUNE_DIR`, `VLLM_SPEC_ATTN_DEBUG`, `VLLM_VISION_CPU_OFFLOAD_GB`, `VLLM_ENGINE_STALL_SENTINEL_S`); the 32 with a raw read on main keep their literal default string exactly, which corroborates the 29-knob check above for everything except the two cells it flags on purpose.

**Not measured, stated plainly.** No throughput or quality run: nothing here changes a kernel or a default except the two cells above, and the tree identity check is the claim. No boot of `long`, `huge`, int4 or offload profiles on this branch; `verify.sh` confirms the kvarn patches applied.

Decisions a reviewer can reject by name: the registration lives in the reading patch rather than in `speed-knobs-envs`; the two moved knobs; the two default cells; `KVARN_*` deferred to the pin flip.


mhenrichsen:
Merging.

Reviewed without a GPU (it is on a training job) — which is the right way to judge this PR anyway, since nothing here touches a kernel.

What I checked independently of the PR body:

- `git-apply` CI is green on the head, and that job is now the fuzz-0 series check against the pin in `docker/requirements.txt` rather than a hard-coded tag — so the check that matters for this change is the one this change installs.
- The six fuzz-applying patches are the real defect being fixed here. I hit exactly this two days ago from the other direction: `prefill-attn-int8.patch`'s `flash_attn.py` hunk applied at offset 170 with fuzz 2 onto a tree where it was already present, and left the int8-QK block duplicated. Dead code after the first `return output`, so inert — but it is the failure mode you are describing, observed in production's own install rather than in theory. Fuzz-0 apply plus `--no-backup-if-mismatch` would have made it a loud failure instead of a silent triplicate.

The knob registration is the part I would have asked for separately, so taking it here is a clear gain: 24 names read but never registered means 24 names outside the torch.compile cache key, which is #114 and then some. The positive control (one made-up `VLLM_` name producing exactly one warning line) is the right shape of evidence for that claim.

Taking the decisions you listed by name as offered: registration in the reading patch, the two moved knobs, the two intentional default cells, `KVARN_*` deferred to the pin flip.

One follow-up that is on me, not on this PR: the deploy tree on the box is a separate copy that drifts, and it is currently 12 of 40 patches behind main. Regenerated patch files mean it needs a full re-apply from this series rather than an incremental one. I will do that when the GPU frees up, and `PATCHES.md` makes that job legible for the first time.

PR #120, opened 2026-09-16
prepare: quant_heads_stream handles lm_head and embed_tokens in different shards

kinggongzilla:
Code and PR description written by qwen3.8 27b:

**Problem**

The head-requant pass assumed `lm_head` and `embed_tokens` live in the same shard and asserts otherwise. That holds for the single-shard export it was written for, but multi-shard asymmetric-AWQ exports place the two heads in different shards (typically first and last), and the script dies at the assert before touching anything on exactly the checkpoint class its docstring claims to cover.

**Fix**

Group the two head keys by shard and run one streaming pass per shard. The same-shard layout collapses to the original single pass (behavior unchanged); the split layout rewrites each shard with the same raw-byte streaming copy, the same `.bak-orig` backups, and the same symmetric group-128 int8 outputs (fp16 scales for `lm_head`, bf16 for embeddings), updating the index per shard. The MTP pass and the config-group fix are untouched.

**Verification**

Reproduced on a 7-shard asymmetric-AWQ export (heads in shards 1 and 7, MTP in a separate small shard): old code dies at the assert; the patched code requantizes both heads and the MTP module in place, and the result passes `verify.sh --no-server`.


mhenrichsen:
Merging.

Checked the two paths on the 3090 with a synthetic export rather than taking the assert at its word — heads only, CPU, no GPU (it is on a training job). Same fixture, `prepare/quant_heads_stream.py` from main and from this branch:

- **same shard** (`lm_head` and `embed_tokens` both in `s1`): main and this branch behave identically — one streaming pass, `rewriting s1.safetensors (streaming)`, then both run on into the MTP stage together. Your claim that the single-shard layout collapses to the original pass holds.
- **split shards** (`lm_head` in `s1`, `embed_tokens` in `s7`, the 1-and-7 layout you describe): main dies at `AssertionError: lm_head and embed_tokens live in different shards` before touching anything. This branch gets past it and rewrites both — `rewriting s7.safetensors (streaming)` after `s1`.

(My fixture's MTP section was incomplete, so both trees stop at the same `KeyError` in the MTP stage afterwards — identically, in all four cells, so it does not bear on the head pass this PR changes.)

The grouping is the right shape: `groups.setdefault(wm[key], []).append(...)` makes the one-shard case a single group by construction rather than by a special case, so there is no second code path to drift. Per-shard `.bak-orig` and per-shard index update both follow correctly.

Noting for the record that the code and the description are qwen3.8-27b's — the fix is correct and minimal, and it is the checkpoint class the docstring already claimed to cover, so the bug was real.

PR #122, opened 2026-09-16
verify: accept the int4-GPTQ lm_head layout the drafter pipeline produces

liamwh:
## What

`verify.sh` hard-required `config_groups.group_1` (lm_head) to be `num_bits == 8`. Pointing `MODEL` at a fast-variant dir — the exact layout `drafter/gptq_lm_head.py` writes and `prepare/fetch_fast_variant.py` distributes — therefore always failed with `lm_head not requantized`, including for the repo's own shipped fast variant.

This accepts 4 or 8, and additionally checks the packed geometry implied by the declared width against the shard headers, so a config claiming one width over the other's tensors still fails (e.g. int8 declared over int4 tensors: packed `248320x640` vs implied `248320x1280`).

## Validation

Ran the model-check heredoc directly against real model dirs:

- int8 dir (`prepare/quant_lm_head.py` output): PASS before and after
- int4 fast-variant dir: FAIL before, PASS after
- fast-variant dir with config patched to claim int8: FAIL on the packed-shape mismatch

(Context: I build target-calibrated fast variants for third-party Qwen3.8-family checkpoints on this stack; this false rejection is why my serving setup ran with `VERIFY=0` for the fast dir until I had a layout-aware check. Happy to adjust to whatever shape you prefer — the geometry check could also extend to embed/MTP, I kept it to lm_head to stay minimal.)


mhenrichsen:
Merging, and taking your version over #126's, which fixes the same false rejection more loosely. The geometry check is the reason.

Ran all four cells on the 3090's real model dirs — header reads only, no GPU, it is on a training job. The model-check block extracted from each branch, run against each dir:

| block | base `…-AutoRound` | fast `…-AutoRound-fast` |
|---|---|---|
| main | PASS | **FAIL** `lm_head not requantized` |
| this PR | PASS int8, geometry matches | PASS int4, geometry matches |
| #126 | PASS 8 bits | PASS 4 bits |

So main rejects the repo's own shipped fast variant, which is #125 and the reason your setup ran with `VERIFY=0`.

Then the case that separates the two fixes. I copied the fast dir's config, flipped the lm_head group to `num_bits: 8`, and pointed both blocks at it — a config lying about its own width over unchanged int4 tensors:

- this PR: `FAIL lm_head declares int8 but packed geometry {'lm_head.weight_packed': (248320, 640), …} != implied {'lm_head.weight_packed': (248320, 1280), …}`
- #126: `PASS lm_head quantized to 8 bits`

`verify.sh` exists to refuse to serve a model that will fail strangely later. A check that reads the declared width and then does not look at what is on disk has given up the thing it was for; yours is the one that keeps it. `V * bits // 32` also reproduces both real layouts exactly, which is a decent argument that the formula is right and not just fitted to one of them.

On extending it to embed/MTP — keep it minimal as you have it. The lm_head is where the two shipped layouts actually differ; the rest can follow a real failure rather than precede it.

Thanks for the context on building target-calibrated fast variants for third-party checkpoints. That is a use of this stack I did not know was happening, and `VERIFY=0` as the workaround is a bad outcome we should not have made you reach for.

PR #123, opened 2026-09-16
verify: fail on tensors present in more than their mapped shard

liamwh:
## What

A whole-directory duplicate-tensor scan for the model check: read every `.safetensors` header in the dir (not `.bak` files) and FAIL when an index-mapped name's holders != [its mapped shard].

## Why

vLLM reads every key of every shard it opens, not just the index-mapped ones. A tensor that lives in more than its mapped shard gets loaded twice — the second load wins, or a shape mismatch aborts the boot far from the cause — while every existence check passes. This is not hypothetical: assembling a model dir by hardlinking shards from the dir it was built from leaves superseded tensors behind when a later stage rewrites one shared file. Concretely, an int8 MTP module surviving inside a hardlinked `model-nonquant` shard loads over the int4 one that replaced it and dies on a have/got shape assert that names neither file. I hit exactly this building a fast variant on top of a prepared dir; it took a while to find precisely because nothing else flagged it.

## Validation

- Clean model dir: `every index tensor lives in exactly its mapped shard` PASSes
- A dir with `mtp.fc.{weight_packed,weight_scale,weight_shape}` additionally present in a second shard: FAILs naming all three and both files

Header-only reads, stdlib `struct`+`json`, no model loading. Independent of #122.


mhenrichsen:
Merging.

Reproduced your negative case and, more importantly, checked for false positives — that is the risk with a whole-directory scan, since it reads files the index never mentions.

Ran the model-check block from this branch against the two real dirs on the 3090 (header reads only, no GPU — it is on a training job):

- base `Qwen3.8-27B-W4A16-AutoRound`: `PASS every index tensor lives in exactly its mapped shard (no duplicates across 8 mapped + stray files)`
- fast variant `…-AutoRound-fast`: same PASS

So the scan is clean on both shipped layouts, including the fast variant's extra files, and the `".bak" not in f` filter correctly skips the `.bak-orig` shards `prepare` leaves behind.

The failure mode you describe is worth the 22 lines on its own. "The second load wins, or a shape mismatch aborts the boot far from the cause" is exactly the kind of thing that costs a day, and every existence check passing while it happens is why. The hardlink-assembly case is not exotic either — that is how anyone builds a variant on top of a prepared dir.

Independent of #122 as you say; both are landing.

PR #124, opened 2026-09-16
Translate chat-template effort vocabulary so OpenAI clients do not get HTTP 400

TyroneNel:
Rebased onto current `upstream/main` (`d1df6dc`), so this is now one commit and four files: the new script, one guarded step in `docker/prepare.sh`, Gotcha 58, and the knob lines in `docs/docker.md`.

## Change
`prepare/translate_chat_template.py` rewrites the shipped effort block in place:

- it maps **only** the names the template does not know — `minimal` -> `low`, `high`/`max` -> `xhigh`;
- every other value **falls through unchanged**, so the template's own levels (`low`/`medium`/`xhigh`) behave exactly as they do today and an omitted `reasoning_effort` keeps the template default (`xhigh`);
- the `raise` is dropped, so an unknown value no longer 400s. It yields no reasoning instruction — the same outcome as `medium` — and nothing is silently rewritten to a stronger level.

`docker/prepare.sh` gains one guarded step (`TRANSLATE_EFFORT=0` skips it) beside the `harden_chat_template.py` step from #116, passing the prepared directories (`BASE`, `BASE-fast`, `MODEL`) as **array** arguments so a path containing a space cannot split.

`docs/docker.md` names both chat-template knobs (`HARDEN_TEMPLATES=0`, `TRANSLATE_EFFORT=0`) where the prepare steps are listed: the repo documents its knobs in prose there (`FAST_VARIANT=0`, `DFLASH2=0`, `PREPARE=0`/`VERIFY=0`), and `HARDEN_TEMPLATES` had landed in #116 undocumented. Docs only — no behaviour changed after the verification below was run.

## Review changes applied
- **The default is `xhigh` again**, and the `docs/quality.md` edit is dropped — no measured baseline moves.
- **The `flock` hunk is split out** into #134, so this PR is only the template translation.
- **`$DIRS` -> bash array**, not quotes: `$DIRS` was a multi-directory string, so `"$DIRS"` would have passed it as one argument. A directory name containing a space is covered in the fixture.
- **An unrecognised effort block warns and exits 0** rather than failing prepare, matching #116's rule — this runs under `set -e` after the ~19.5 GB download and requantisation.
- **Gotcha 58 stays**, with its `Fix:` paragraph rewritten to the new semantics.

## v1 -> v2 upgrade (found while rebasing)
The first cut wrote a `v1` marker with the `medium` default, and the marker short-circuits re-runs — so every already-prepared model dir would have kept `medium` after this change. The script now recognises the v1 block and upgrades it (reported as `upgraded`). The `.jinja.bak-effort` backup is no longer written.

## Verification (CPU only; no model dir is written)
The real template rendered through `transformers.apply_chat_template` — the path vLLM uses — against copies:

| input | pristine | after |
|---|---|---|
| omitted | xhigh instruction | **byte-identical** |
| `low` / `medium` / `xhigh` | low / none / xhigh | **byte-identical** |
| `minimal` | `TemplateError: Unexpected reasoning effort minimal` (the reported 400) | renders, low instruction |
| `high`, `max` | `TemplateError` | renders, xhigh instruction |
| `none`, `banana` | `TemplateError` | renders, no instruction (documented) |

Also: pristine -> `translated`; second run -> `already`; a v1 file -> `upgraded`, then matches pristine on all four unmapped inputs and is idempotent; foreign shape -> `WARN`, exit 0, file byte-identical; a directory containing a space -> `translated` with the same output as the plain path; a missing dir -> `missing`; `bash -n docker/prepare.sh` and `py_compile` clean.

## Limits
No image rebuild and no full `prepare` run (~19.5 GB re-download). No live vLLM request — the render path is transformers', not a running server. One open question for the reviewer: `none` currently yields no reasoning instruction (like `medium`) rather than mapping to `low`; `minimal` and `none` both mean "least effort", so mapping it may be more faithful, but I left it out of the map because it was not in the list.


mhenrichsen:
The bug is real and the translation approach is right. Two things to change before this can go in, one of them a behaviour change I do not think you intended to be the headline.

**1. The default must stay `xhigh`.** The script sets the template default to `medium`, and `docs/quality.md` is edited to match. That is not a client-compatibility fix — it silently downgrades every request that sends no `reasoning_effort` at all, which is most of them. Every quality number in this repo (GSM8K 98.3, the PPL series, the acceptance figures) was measured at the shipped `xhigh` default; changing it re-baselines all of them with no measurement offered, and on a box where the GPU is currently on a training job I cannot re-measure to tell you what it costs.

Mapping the vocabulary and changing the default are separable, and only the first is needed for the 400s: keep `xhigh` as the default, map `minimal` → `low`, `high` and `max` → `xhigh`, and fall back to the template default rather than to `medium` on an unknown value. If you want `medium` as the default, that is a real proposal, but it wants its own PR and a quality run behind it.

**2. The `flock` hunk is undeclared scope.** The body lists the translate script, a 15-line `prepare.sh` hook, and docs. `docker/prepare.sh` also gains:

```
exec 9>"${BASE_MODEL_DIR:-/app/models}/.prepare.lock"
flock -n 9 || { echo "prepare: another preparation holds …"; exit 1; }
```

Serialising prepare may well be a good idea — but it is a new failure mode for every user (a stale lock now refuses to prepare), it is unrelated to chat templates, and it is not in the description, so nobody reviewing the stated change would look at it. The lock path is also inconsistent: with `BASE_MODEL_DIR` set it lands *inside* the model directory, unset it lands in `models/`. Split it out and I will review it on its merits.

**Smaller notes.**

- `python prepare/translate_chat_template.py $DIRS` — unquoted, so a model path containing a space splits. #126 deliberately covers that case in its selection fixture; worth not regressing it here.
- This overlaps #116, which merged a few minutes ago and inserts its own step at the same point in `prepare.sh`. You will need a rebase regardless. The two are complementary — #116 hardens array tool-call arguments, this translates the effort vocabulary — and both now sit after the same `done`.
- #116 landed with the rule that an unrecognised template **warns** rather than failing prepare, for the same reason it matters here: this runs under `set -e` after the ~19.5 GB download and the requantisation. Your script "stops if the effort block has a new shape that it does not know" — make that a warning too, or the first upstream template revision turns into a failed prepare over a ready model dir.
- Gotcha 58 is well written and I want it in the docs whatever happens to the rest.

Reproduction details in the body are good — the exact `jinja2.exceptions.TemplateError`, the template line, the vLLM call path, the date you saw it in service. That is what made this quick to judge.

TyroneNel:
Thanks — rebased onto current `main` (`d1df6dc`) and applied every point. All four of the smaller notes are in as well.

- **The default is `xhigh` again.** The map is now only what you listed and nothing else: `minimal` -> `low`, `high`/`max` -> `xhigh`. Every other value falls through unchanged, so `low`/`medium`/`xhigh` behave exactly as the template does today and an omitted `reasoning_effort` keeps the `xhigh` default. The `medium` default and the `medium` fallback are gone, and the `docs/quality.md` edit is dropped — no baseline moves.
- **`flock` split out** into #134. Its lock path now comes from `dirname "$BASE"`, so it no longer lands *inside* the model dir when `BASE_MODEL_DIR` is set, and the wait is bounded by `PREPARE_LOCK_WAIT` (default 600 s) instead of refusing on sight.
- **`$DIRS`** is fixed with a bash array rather than quotes — `$DIRS` was a multi-directory string, so `"$DIRS"` would have passed one argument and lost the rest. The fixture covers a directory name with a space.
- **Warn, don't fail**: an unrecognised effort block prints `WARN` and exits 0, and a re-run is a `already` no-op rather than a failed prepare.
- **Gotcha 58 stays**, with its `Fix:` paragraph rewritten to the new semantics.

One thing worth flagging, because your review is what surfaced it: the marker. It was `effort-translation v1` and it short-circuits re-runs, so every dir already translated by the first cut would have silently kept the `medium` default after this change — the exact regression you were pushing back on, surviving the fix. The script now recognises the v1 block and upgrades it in place, and I dropped the `.jinja.bak-effort` backup.

Evidence is in the body: the real template rendered through `transformers.apply_chat_template`, unmapped values byte-identical to pristine, `minimal`/`high`/`max` flipping from the `TemplateError` you reproduced to rendered output, plus the v1 -> v2 upgrade and the space-in-path case. One open question: `none` yields no reasoning instruction (same as `medium`) rather than mapping to `low` — I kept it out of the map since you did not list it, but say the word if you would rather it map.

TyroneNel:
Added the knob documentation to `docs/docker.md`, and named the knob in gotcha 58's `Fix:` paragraph. Four files now: the script, the `prepare.sh` step, and the two docs.

On scope, since I got this wrong the first time: this is deliberately not the `flock` situation. No behaviour changes, and the knob is one this PR introduces. My first cut left it only in `prepare.sh` and this description, which meant the repo had no record of it outside a code comment. `docs/docker.md` already lists the prepare knobs in prose (`FAST_VARIANT=0`, `DFLASH2=0`, `PREPARE=0`/`VERIFY=0`), so the line belongs with them.

I also named `HARDEN_TEMPLATES=0` in the same sentence: it is #116's knob, it landed undocumented, and it drives the step that sits directly beside this one. If you would rather see that fixed separately, drop the clause — it is four words.

The doc delta is doc-only: `git diff` against the revision you last saw touches `docs/docker.md` and `docs/gotchas.md` and nothing else, so the verification in the description still stands unchanged. Head is `cd6dbb9`.

mhenrichsen:
Merging. The default question is settled, and I checked it the way it needed checking rather than reading the diff.

Rendered the real `chat_template.jinja` from the box's model dir through jinja2, pristine against translated, same messages, every effort value:

| effort | pristine | translated | identical output |
|---|---|---|---|
| *(omitted)* | OK | OK | **yes** |
| `xhigh` | OK | OK | **yes** |
| `medium` | OK | OK | **yes** |
| `low` | OK | OK | **yes** |
| `minimal` | `TemplateError` | OK → low | — |
| `high` | `TemplateError` | OK → xhigh | — |
| `max` | `TemplateError` | OK → xhigh | — |

Byte-identical on all four values the stock template supports, so no quality baseline moves — which was the whole of my objection. The three that 400'd now render. `{%- set requested_effort = … else 'xhigh' %}` keeps the default where it was.

Also confirmed on the box: `WARN` + exit 0 on an unrecognised effort block, `already` on a re-run, and a directory path containing a space handled (the bash array, which is the right fix — `"$DIRS"` would have passed one argument and swallowed the rest).

**The v1 marker catch is the best thing in this PR.** A fix that leaves every already-translated directory pinned to the old default would have shipped the exact regression the review was about, invisibly, on the machines that had already run it. Upgrading the v1 block in place is right, and I would not have caught it from the diff.

One behaviour change worth naming, since it is outside what either of us set out to change: `banana` used to raise and now renders with no reasoning instruction. Removing `raise_exception` is the point of the PR, so this follows — but it does mean a client typo is now silently ignored rather than reported. I would not add validation back for it. Worth a line in gotcha 58 if you are touching it again.

On `none`: leave it out of the map, as you have it. It renders like `medium` today and nobody has reported it.

`HARDEN_TEMPLATES=0` documented in the same sentence — keep the clause. It landed undocumented in #116, which was mine.

Issue #113, opened 2026-09-15
bench/warmup.sh duplicates start_qwen.sh's MODEL and API-key resolution


mhenrichsen:
Follow-up from #55, not a blocker for it.

`bench/warmup.sh` re-derives two things that `single-user/start_qwen.sh` already resolves:

* the model dir (`models/Qwen3.8-27B-W4A16-AutoRound`, preferring the `-fast` variant when present)
* the API key precedence chain (`OPENAI_API_KEY` -> `VLLM_API_KEY` -> `api_key.txt` -> placeholder)

They agree today — that was checked when #55 landed. They will drift, and the failure is quiet: a mismatched key 401s every bench request, and a missing model path fails the warmup. Since #55 the warmup is advisory (logs and serves anyway), so drift would show up as a warmup that silently never runs rather than as a broken boot.

Worth extracting a small sourceable resolver that both scripts use. Low priority.

PR #136, opened 2026-09-17
bench: warm up against the model the launcher actually serves (#113)

TyroneNel:
Closes the model-selection half of #113. The key half is a separate PR.

## Why
`bench/warmup.sh` re-derived the model directory itself — the fast variant when present, else the base dir — under a comment that said it mirrors `single-user/start_qwen.sh`. #126 extracts exactly that resolution into `single-user/select_model.sh` and points the launcher and the Docker gate at it, which leaves this copy as the only one that can drift: the warmup would then exercise a different directory than the one being served.

## What
The inline block becomes `source "$REPO/single-user/select_model.sh"` — 8 insertions, 11 deletions. The comment now says what the resolver does (an explicit `$MODEL` wins, else the fast variant when present, else the base dir) rather than describing a copy of the launcher.

## Stacked on #126
The base branch is `fix/f1-fast-model-verifier`, because `single-user/select_model.sh` does not exist on `main` yet — without it the warmup would source a missing file. Merge #126 first; this diff then collapses to the single commit `091e3da`. Opened now so the dependency is visible rather than parked on a local branch.

## Verification
The real `bench/warmup.sh`, run against a fake repo with a stub `curl` (so the health check passes) and a stub `venv/bin/vllm` that records its argv, asserting the `--model` the script actually passes. 11/11 checks (5 cases + 6 equivalence cells), 0 failures:

| case | result |
|---|---|
| base + fast present, `MODEL` unset | ok — passes the fast dir |
| base only, `MODEL` unset | ok — passes the base dir |
| explicit `MODEL`, fast present | ok — preserved |
| repo path containing a space | ok — resolved intact, one argument |
| `MODEL` unset under `set -euo pipefail` | ok — no unbound-variable failure |

Equivalence with the block being replaced, over the same fixtures (fast/base × `MODEL` unset/empty/explicit): the sourced resolver and the inline block produce identical results in all six cells, so which directory gets warmed does not change.

`bash -n bench/warmup.sh` clean.

## Limits
No live server: the two warmup passes are exercised through a stub, so their HTTP content is not checked here. The warmup remains advisory — a failed warmup logs and serves anyway.


TyroneNel:
Rebased on top of the rebased #126 and collapsed to one commit, as you asked in the #126 merge note — `57fdf0f`, `bench/warmup.sh` only, 8 insertions / 11 deletions. Both PRs are `MERGEABLE / CLEAN` now.

It still has to stay stacked until #126 lands, since `single-user/select_model.sh` does not exist on `main` yet and the warmup would source a missing file.

Equivalence re-checked on the rebased tip rather than carried over — old inline block against the sourced resolver, six cells, all identical:

| fast variant | `MODEL` | resolved |
|---|---|---|
| present | empty / unset | `models/…-AutoRound-fast` |
| present | explicit | the explicit dir |
| absent | empty / unset | `models/…-AutoRound` |
| absent | explicit | the explicit dir |

`bash -n` clean.

mhenrichsen:
Same verification as on #126 — this branch merged onto today's `main` is what I ran, since it contains both of #126's commits: fixture suite OK, `bash -n` clean on all five scripts, and `select_model.sh` resolving to the fast directory on the box's real model tree while keeping an explicit `MODEL`. `bench/warmup.sh` here also resolves the served model through the same selector, which is the #113 half.

Held for the same reason as #126 — the `workflow` scope on my token, needed because the branch edits `.github/workflows/patch-integrity.yml`. Merge order once that is fixed: #126, then this.

mhenrichsen:
Merged on top of #126 (which carried this branch's first two commits) and the vLLM 0.29 pin; verified together with #126 on the reference box before either went in.

PR #137, opened 2026-09-17
keys: one precedence resolver for the launchers and the warmup (#113)

### LINKED ISSUE #113: bench/warmup.sh duplicates start_qwen.sh's MODEL and API-key resolution

TyroneNel:
Closes #113.

## Why
`bench/warmup.sh` re-derived the key chain `single-user/start_qwen.sh` already resolves, and the failure is quiet: a mismatched key 401s every bench request, and since #55 the warmup is advisory, so drift would surface as a warmup that silently never runs rather than as a broken boot.

## What
`resolve_api_key.sh` (new, repo root beside the launcher scripts) defines two functions and **sets nothing by itself**, so each caller opts into the side it needs:

- `resolve_vllm_key` — server side. vLLM binds `--api-key` from `VLLM_API_KEY`; `api_key.txt` is the documented file fallback and an exported `VLLM_API_KEY` wins. No placeholder: with no key anywhere the server must bind none, and inventing one here would make it demand a key nobody configured.
- `resolve_client_key` — client side. An explicit `OPENAI_API_KEY` wins (the client may point at a server on another host, where the local file is the wrong key), else the server key, else `api_key.txt`, else the `EMPTY` placeholder a server that bound no key ignores. It reads the server side without exporting it.

Call sites: `bench/warmup.sh` (its 8-line block), `single-user/start_qwen.sh` and `batch/start_qwen.sh` (their identical 2-line blocks). `resolve_api_key.sh` +42, `bench/warmup.sh` +5 −15, `single-user/start_qwen.sh` and `batch/start_qwen.sh` +2 −3 each.

## No behaviour change, checked as equivalence
Both functions are the existing rules moved, so the verification asserts the move rather than the intent.

**8-cell matrix** — `VLLM_API_KEY` set/unset × `api_key.txt` present/absent × `OPENAI_API_KEY` set/unset — running the old inline logic and the sourced resolver in separate processes per cell and comparing both resulting variables: **16/16 identical**, at the launcher call site and at the warmup call site.

**End-to-end** — the real `bench/warmup.sh` against a fake repo (stub `curl` so the health check passes, stub `venv/bin/vllm` dumping the environment it receives):

| fixture | `vllm bench serve` sees |
|---|---|
| key in `api_key.txt` | `OPENAI_API_KEY=FILEKEY` |
| explicit `OPENAI_API_KEY` | `OPENAI_API_KEY=EXPLICIT` |
| caller exported `VLLM_API_KEY` | `OPENAI_API_KEY=ENVKEY` |
| no key anywhere | `OPENAI_API_KEY=EMPTY` |

and in every case `VLLM_API_KEY` is left exactly as the caller set it — the warmup process gains no export it did not have (the client path does not call the server path).

`bash -n` clean on all four files.

## Not in scope, deliberately
- **Four `bench/*.sh` metrics scripts** (`measure_c1.sh`, `prefill_ab.sh`, `real_rep.sh`, `run_benchmarks.sh`) carry `export OPENAI_API_KEY=${VLLM_API_KEY:-$(cat "$REPO/api_key.txt" 2>/dev/null)}`. Their precedence is different on purpose: they always target `127.0.0.1`, so local metrics auth must present the key the **server** bound, and an explicit `OPENAI_API_KEY` is not an override there. Migrating them means choosing a rule per script — say the word and I will do it as its own change.
- **`bench/*.py`** read `api_key.txt` in Python; a shell resolver cannot serve them.
- **CI coverage:** the shell-syntax step globs `bench/ docker/ single-user/ batch/ patches/ verify.sh docker/entrypoint.sh`, so this new root-level script is not syntax-checked there. `bash -n resolve_api_key.sh` is in the verification above; extending the glob is a one-line change I left out of the diff.

## Limits
No server was started and no request was made. The launchers were not run end-to-end (they start a server); their key lines are covered by the matrix above, not by a live boot.


mhenrichsen:
Merging. Ran the 8-cell matrix myself on the box, both resolved variables per cell, each in its own process:

| `VLLM_API_KEY` | `api_key.txt` | `OPENAI_API_KEY` | server | client |
|---|---|---|---|---|
| unset | absent | unset | *(none)* | `EMPTY` |
| unset | absent | `EXPLICIT` | *(none)* | `EXPLICIT` |
| unset | present | unset | `FILEKEY` | `FILEKEY` |
| unset | present | `EXPLICIT` | `FILEKEY` | `EXPLICIT` |
| `ENVKEY` | absent | unset | `ENVKEY` | `ENVKEY` |
| `ENVKEY` | absent | `EXPLICIT` | `ENVKEY` | `EXPLICIT` |
| `ENVKEY` | present | unset | `ENVKEY` | `ENVKEY` |
| `ENVKEY` | present | `EXPLICIT` | `ENVKEY` | `EXPLICIT` |

Exactly the two rules as stated, and the two that matter are right: **no key anywhere leaves the server binding none** rather than inventing a placeholder that would make it demand a key nobody configured; and an explicit `OPENAI_API_KEY` always wins on the client side, which is the case where the client is pointed at another host and the local file is the wrong key.

Splitting into two functions that set nothing by themselves is the right shape — the warmup needs the client side without acquiring a server-side export it did not have.

Agreed on both exclusions. The four `bench/*.sh` metrics scripts genuinely have different precedence — they always target `127.0.0.1`, so they must present what the server bound and an explicit `OPENAI_API_KEY` is not an override there. Do not migrate them to look uniform; they are not.

**One thing to fix, and it applies to #138 too:** the CI shell-syntax job globs `bench/ docker/ single-user/ batch/ patches/ verify.sh docker/entrypoint.sh`, so `resolve_api_key.sh` at the repo root is not syntax-checked. You flagged it and left it out of the diff, which was right for scope — I will extend the glob myself once #138 lands, so both new root scripts are covered by one change.

PR #138, opened 2026-09-17
launchers: one validated effective-configuration resolver (F13 / backlog 6)

TyroneNel:
One validated effective-configuration resolver for the launchers (F13 / automation backlog item 6).

## Why
The launchers resolved `CTX`/`SPEC`/`KV` themselves, so a typo'd profile could reach the boot command and the failure would surface as a wrong geometry rather than as a refusal. `single-user/start_qwen.sh` also ended its `CTX` chain in a bare `else`, so *any* unrecognised value silently took the `long` arm.

## What
`resolve_config.sh` (new, repo root) is sourced by both launchers right after `REPO` is set. It:

1. **Refuses** (exit 1) unknown values for the enumerated controls — `CTX` and `SPEC` in single mode, `KV` in batch mode — with the accepted set in the message.
2. **Warns** (stderr, continues) about controls that are set but silently ignored in that mode (`CTX`/`SPEC` in batch, `KV` in single) and about launcher flags shadowed by `EXTRA_ARGS`, which expands last and wins without a word.
3. **Prints the redacted effective configuration** to stderr — every control with its resolved value, secrets presence-only (`VLLM_API_KEY` as `set(N chars)`, never its value; `api_key.txt` as present/absent).

Because validation now owns the refusal, the silent `else` in `single-user/start_qwen.sh` becomes `elif [ "$CTX" = "long" ]`, so every remaining arm is reachable.

**One addition beyond the original local commit:** the `source` line is guarded in both launchers:

```bash
source "$REPO/resolve_config.sh" \
  || { echo "start_qwen: cannot source $REPO/resolve_config.sh - refusing to boot unvalidated" >&2; exit 1; }
```

Neither launcher runs under `set -e`, so a missing file would otherwise skip the check entirely and boot unvalidated — the inverse of what this file is for. I found that by accident while building the fixture below, and it is the one thing here that changes behaviour rather than moving it.

## Verification
29 checks, 0 failures, all through the file's own interface:

- **Refusals** — `CTX=bogus`, `SPEC=bogus`, `KV=bogus`, an unknown mode, and an unset `REPO` all exit 1; the message names the control and the accepted set.
- **Accepted profiles and defaults** — `CTX=long SPEC=dflash2`, `KV=kvarn`, and the no-`env` case (single → `CTX=fast SPEC=mtp`, batch → `KV=fp8`); all exit 0 and print the effective config.
- **Warnings** — `KV` set in single mode, `CTX`/`SPEC` set in batch mode, an `EXTRA_ARGS` flag that shadows a launcher flag, and an inline `--api-key`; the inline key value is not echoed.
- **Redaction** — with `VLLM_API_KEY=SUPERSECRETVALUE` and a populated `api_key.txt`: `VLLM_API_KEY=set(16 chars)`, `api_key.txt=present`, and neither the key value nor the file contents appear anywhere in the output.
- **Fail-closed wired into a launcher** — the wiring lines are extracted **verbatim from `single-user/start_qwen.sh`** and run in a stub context, so the test cannot drift from the file: `CTX=bogus` stops the launcher (exit 1, no boot), `CTX=long` lets it through, and with `resolve_config.sh` removed it refuses with the message above instead of booting.
- `bash -n` clean on `resolve_config.sh` and both launchers.

## Limits
No launcher was booted end-to-end — they start a server — so the wiring is exercised in a stub context using the launcher's own lines. No GPU work.


mhenrichsen:
Merging. Exercised it through its own interface on the box — `bash resolve_config.sh batch|single`, `env -i` so nothing leaks in from my shell:

| invocation | exit | output |
|---|---|---|
| `single CTX=bogus` | 1 | `refusing: unknown CTX=bogus (want fast\|long\|huge)` |
| `single SPEC=bogus` | 1 | `refusing: unknown SPEC=bogus (want mtp\|dflash2\|off\|none)` |
| `batch KV=bogus` | 1 | `refusing: unknown KV=bogus (want fp8\|kvarn\|int4pth)` |
| unknown mode | 1 | `refusing: unknown mode nonsense (want batch\|single)` |
| `REPO` unset | 1 | `refusing: REPO is unset` |
| `single CTX=long SPEC=dflash2` | 0 | effective config printed |
| `batch KV=kvarn` | 0 | effective config printed |
| `single` / `batch` (no env) | 0 | `CTX=fast SPEC=mtp` / `KV=fp8` |
| `single KV=kvarn` | 0 | `WARNING: KV=kvarn is set but single-user mode ignores it` |
| `batch CTX=long` | 0 | `WARNING: CTX=long is set but batch mode ignores it` |

Redaction with `VLLM_API_KEY=SUPERSECRETVALUE` and a populated key file: `VLLM_API_KEY=set(16 chars) api_key.txt=absent`, and the value appears nowhere in the output. Refusals exit 1 for real, which is what makes the fail-closed claim true rather than aspirational.

**The bare `else` is the actual bug here** and it is worth saying out loud: `CTX` fell through to the `long` arm for *any* unrecognised value, so a typo did not fail — it booted 150k context with fp8 KV and FlashInfer, and you would only find out from the geometry in the log. Turning it into `elif [ "$CTX" = "long" ]` once validation owns the refusal is the right order of operations.

**The guarded `source` is the right call too.** Neither launcher runs under `set -e`, so an unsourceable validator would have skipped validation silently — a validator that fails open is worse than none, because it is trusted. Good catch, and good that you flagged it as the one behaviour change rather than burying it.

The warning about `EXTRA_ARGS` shadowing earns its place: it expands last and wins without a word, and that has cost time in at least two issues.

Two follow-ups, both mine:

- I am extending the CI shell-syntax glob to cover the repo-root scripts, so this and `resolve_api_key.sh` from #137 are checked. Neither is today.
- #126 also touches `single-user/start_qwen.sh` and is waiting on a permission detail on my side. If it conflicts when it lands, I will rebase it rather than send it back.

Issue #62, opened 2026-09-02
Field report: RTX 3090 / WSL2 — int8 prefill stack confirmed (+59% / +6.3%), plus a power-limit caveat for 3090 benchmarks


mjungnickel18:
## Setup

RTX 3090 (24 GB, sm_86), Windows 11 + WSL2/Ubuntu, venv install (not Docker), repo at `bb739e4`, `verify.sh --install` clean. Turbo profile: `CTX=fast SPEC=dflash2 MAX_SEQS=2 KV_MEM=6400000000 DFLASH_MAX_LEN=79872`, W4A16 AutoRound checkpoint.

Prefill measured as TTFT with `max_tokens=1`, token counts taken from `usage.prompt_tokens` (not estimated), German prose input, each repetition prefixed with a unique header so prefix caching cannot fire — verified via `usage.prompt_tokens_details.cached_tokens == 0`. **Power limit set after service start and re-verified at 250 W before every cell**, medians, n=5 (n=3 at 66k).

## Results

| prefill tok/s | 5,216 tok | 20,729 tok | 66,093 tok | decode (code, 400 tok) |
|---|---:|---:|---:|---:|
| stock | 1,131 | 1,048 | 851 | 91.6 |
| `INT8_ACT=int8` | **1,802** (+59%) | **1,645** (+57%) | **1,166** (+37%) | 97.9 |
| `+ PREFILL_ATTN=int8` | 1,806 (+0.2%) | 1,674 (+1.8%) | **1,239** (+6.3%) | 106.5 |

SD across repetitions was 1.2–1.6% on the clean cells.

**The int8-QK kernel's context scaling reproduces your own numbers closely** — you state +2.7% at 16k and +5.3% at 51k; we measure +1.8% at 21k and +6.3% at 66k. Nothing to fix here, just an independent confirmation on Ampere/WSL2.

**A warning worth putting in the docs, though:** we first tested `PREFILL_ATTN=int8` *without* `INT8_ACT`, and measured it 1–6% **slower** than stock at every length. The `start_qwen.sh` comment does say "on top of INT8_ACT", but it is easy to read `PREFILL_ATTN` as a standalone switch — and on its own it is a small regression, not a small win. Might be worth an explicit "requires INT8_ACT to pay off" line next to the flag.

## Unrelated finding that may matter more for 3090 owners

[...]

PR #139, opened 2026-09-17
docs: third-party entry for a fully-built Swift fast variant

liamwh:
Adds an entry to `docs/third-party-checkpoints.md` for
[liamwh/Swift-Qwen3.8-27B-W4A16-syv-fast](https://huggingface.co/liamwh/Swift-Qwen3.8-27B-W4A16-syv-fast):
[ukisai/Swift-Qwen3.8-27b](https://huggingface.co/ukisai/Swift-Qwen3.8-27b)
(the reduced-reasoning finetune) published in the complete fast-variant
layout — TheUnderscore's AWQ body carried through, int4-GPTQ lm_head/MTP
calibrated on the finetune's own hidden states, and a draft vocab counted
over its own outputs — so it serves with zero preparation, the way
leminkozey's AutoRound build does, but at fast-variant speed.

Why it may be worth an entry beyond "one more checkpoint":

- **A worked example of building the fast-variant layout for a finetune**
  the prebuilt fast variant does not cover, using this repo's `drafter/`
  pipeline end to end (gen_data → capture → GPTQ heads → build_draft_vocab).
- **A result the community keeps asking about with finetunes**: with the
  heads and vocab rebuilt, the DFlash2 acceptance deficit vs the base model
  did not materialise (0.385 vs 0.387 with the generic drafter), and MTP
  acceptance is *higher* than the shipped fast variant's (0.660 vs 0.634,
  98.4 vs 98.2 tok/s at `SPEC=mtp CTX=long`).
- **A vocab finding that generalises**: the finetune emits only ~25.9k
  distinct tokens (base: ~54k), so its draft head is 25,879 rows rather
  than 40,960 — smaller and faster, 99.8% held-out coverage vs 96.7% for
  the shipped id list.
- Two build notes for anyone rebuilding this layout on other exports: the
  upstream AWQ splits lm_head/embed_tokens across shards (the case #37's
  streaming requant was just generalised for), and a stale-tensor
  collision to avoid when assembling the dir with hardlinks (vLLM loads
  every key in an opened shard, so superseded int8 MTP tensors left in a
  kept shard collide with the int4 ones in extras).

The card documents the full recipe, measurements and both licences the
Swift Open License requires a derivative to carry. Happy to trim the entry
if the maintainers prefer the shorter leminkozey-style treatment.


TyroneNel:
Nice work on this. The own-output draft vocab finding (~25.9k distinct tokens, 99.8% coverage) is really useful, and it's good to see the DFlash2 acceptance deficit disappear once the heads and vocab are rebuilt for the finetune.

I've just finished a similar quant of the uncensored checkpoint from orcarouter:

- **Base:** [orcarouter/Qwen3.8-27B-Uncensored](https://huggingface.co/orcarouter/Qwen3.8-27B-Uncensored)
- **Quant:** [TyroneNel/Qwen3.8-27B-Uncensored-W4A16-g128](https://huggingface.co/TyroneNel/Qwen3.8-27B-Uncensored-W4A16-g128)

Launched with:

```bash
MODEL=/models/Qwen3.8-27B-Uncensored-W4A16/Qwen3.8-27B-Uncensored-BF16-w4g128 \
SPEC=dflash2 CTX=long PREFIX_CACHE=1 KV_MEM=4456028569 DFLASH_MAX_LEN=98304 \
bash single-user/start_qwen.sh
```

The difference was night and day. I could leave it running continuously on long jobs, and it kept going until the work was actually done, without stopping early or crashing. I think it's an excellent candidate.

I'm now finishing an uncensored Swift quant to see how the reduced reasoning affects accuracy.

<details>
<summary><b>Quant specs</b> (both dirs)</summary>

| | `Qwen3.8-27B-Uncensored-W4A16` | `Swift-Qwen3.8-27B-Uncensored-W4A16` |
|---|---|---|
| **Architecture** | `Qwen3_5ForConditionalGeneration` (text_config: 64 layers, GQA 24 heads / 4 KV, head_dim 256, hidden 5120, vocab 248,320, MTP head grafted, `tie=False`) | identical |
| **Body quant** | W4A16, symmetric, per-channel, g=128 (pack-quantized, groups 0–3, 206 ignore entries; the 110 vision-tensor entries are intentionally BF16) | identical |
| **lm_head + embed_tokens** | int8, g=128 packed | identical |
| **MTP head (8 linears)** | int8 (rel-err 0.007–0.015) | identical |
| **Draft head** | 40,960-row int8 + `mtp_draft_vocab_ids.pt` | identical |
| **KV-cache scheme in config** | none (serving-time dtype) | identical |
| **Size on disk** | 15.61 GiB (backups cleaned) | 15.61 GiB (backups still present, ~4.6 GiB extra) |
| **Index** | 2022 entries / 67 shards, 0 missing, 0 orphans | 2022 entries / 67 shards, 0 missing, 0 orphans |
| **Quant recipe** | AutoRound: `asym:False`, `gptlmhead:True`, `mtp:True`, vision: disabled | Swift: `asym:False`, `gptlmhead:True`, `mtp:True`, vision: enabled |

</details>

sudoeste:
Tried your model but can't use  dstyle = int8

Edit:

Oh, I see it's asymmetric :/

TyroneNel:
will have a look at that. thanks for the feedback.

mhenrichsen:
Merged.

I checked the artifact's structure independently before merging — HTTP range reads of the safetensors headers plus the index, no download — and every structural claim in the entry holds:

```
2,422 index entries / 2,422 keys over 9 files, 0 duplicated, 0 orphaned, 0 missing
lm_head.weight_packed                  I32 [248320,  640]   int4 (group_1: 4)
model...embed_tokens.weight_packed     I32 [248320, 1280]   int8 (group_2: 8)
mtp.fc.weight_packed                   I32 [  5120, 1280]   int4, extras only
mtp.draft_lm_head.weight_packed        I32 [ 25879,  640]   + its own ids file
```

So the 25,879-row draft head is really there, in int4, at the width the fast-variant layout expects, and the stale-tensor collision your build note warns about is absent from your own dir — which is the thing worth verifying, given that exact collision shipped broken in another checkpoint this week ([#144](https://github.com/syv-ai/HyperQwen/pull/144), now fixed).

**The vocab finding is the part of this entry that outlives the checkpoint.** ~25.9k distinct tokens against base Qwen's ~54k, 99.8% held-out coverage from the rebuilt list against 96.7% from the shipped one, and the DFlash2 acceptance deficit (0.385 vs 0.387) not materialising once heads and vocab are rebuilt — that is the answer to a question that comes up every time someone finetunes this model, and it now has a measurement attached instead of an opinion. It is already doing work: the Swift+uncensored entry merged just before this one ships the base 40,960-id list, and its entry now points here for what rebuilding would buy it.

One caveat I have left as written rather than edited: the 98.4 tok/s is at **350 W**, and everything else in this repo is a 250 W 3090 (`docs/reproductions/README.md` explains why that matters — the 200/250/280 W ladder in #62 moves throughput by a third). It reads correctly as a same-box A/B against your own run of the shipped fast variant at 98.2, and that is how I take it; if either arm came from a different box or power cap, say so and I will annotate the entry.

Thanks for the build notes — the split-shard lm_head/embed_tokens case and the hardlink collision are exactly what the next person rebuilding this layout will hit.

PR #144, opened 2026-09-18
docs: add Swift + uncensored ready-made checkpoint

ultimaterex:
[ultimaterex/Swift-Qwen3.8-27B-Uncensored-W4A16-AutoRound](https://huggingface.co/ultimaterex/Swift-Qwen3.8-27B-Uncensored-W4A16-AutoRound) combines [ukisai/Swift-Qwen3.8-27b](https://huggingface.co/ukisai/Swift-Qwen3.8-27b)'s reasoning-token-efficiency adapter with an abliterated, uncensored base ([d0xin/Swift-Qwen3.8-27B-Uncensored-BF16](https://huggingface.co/d0xin/Swift-Qwen3.8-27B-Uncensored-BF16)), quantized with this repo's own AutoRound + prepare/ recipe — same shape as leminkozey's checkpoint, so it serves without any preparation.

 Fewer reasoning tokens per turn matters most in agentic use, where the model re-reasons on every tool call and that cost compounds. Pairing it with an uncensored base is aimed at the same use case — fewer stalled tool calls from unnecessary refusals, on top of the speed gain.

Validated with a 6-task battery (arithmetic, code generation, factual recall, a constraint-logic puzzle, a security-training explanation, strict output-format compliance) against the base checkpoint on an RTX 3090: 6/6 correct on both, and a measured 31.2% reduction in reasoning tokens on that battery, real, but short of the adapter's own "58.3%" headline claim, and not uniform across tasks.


sudoeste:
Why would you create this pr and gate the access to your hf card?

ultimaterex:
@sudoeste Overlooked that from the base repo I started with, it's ungated now and I've improved performance by ~15-18%. after recalibrating the draft head.

TyroneNel:
Good work, i will definitely test it.
not sure if you checked the [orcarouter/Qwen3.8-27B-Uncensored](https://huggingface.co/orcarouter/Qwen3.8-27B-Uncensored) version. 
overall it is giving me a better experience and reasoning-token-efficiency speed is also up there.
https://github.com/syv-ai/HyperQwen/pull/139#issuecomment-5718513635
let me know if you give it a go i would like to know what the outcome is.

sudoeste:
It seems the int4 mtp reupload has some stale int8 (as in pr 189)

ultimaterex:
> Good work, i will definitely test it. not sure if you checked the [orcarouter/Qwen3.8-27B-Uncensored](https://huggingface.co/orcarouter/Qwen3.8-27B-Uncensored) version. overall it is giving me a better experience and reasoning-token-efficiency speed is also up there. [#139 (comment)](https://github.com/syv-ai/HyperQwen/pull/139#issuecomment-5718513635) let me know if you give it a go i would like to know what the outcome is.

It seems to be doing much better in my benchmarks and long horizon taks, i think the abliteration I picked is causing lower scores, your version scores much better overall in correctness.

TyroneNel:
Did a full run on the three models last night with the repo bench guidance spec. for interest sake. I'm just playing and learning. My goal is to have a fast model that can run teams for run long tasks without crashing and good accuracy. Any suggestions are welcome.

## Results

### Single profile (`CTX=fast`, `SPEC=dflash2`, 8 seats, 64k) — real-prompt cohorts

e2e output tok/s, `T=default` / `T=0` (greedy); `tok/step` = speculative tokens
accepted per step (DFlash2, exact by construction).

| cohort | orca | swift | autoround-base | published (DFlash2) |
|---|---|---|---|---|
| C1 | **121.6** / 128.5 | 119.1 / 123.4 | 119.2 / 124.0 | 121.8 |
| C2 | **185.7** / 194.4 | 183.8 / 182.6 | 185.7 / 193.1 | 195.5 |
| C4 | **242.1** / 255.3 | 224.1 / 239.5 | 234.9 / 248.6 | 278.9 |
| C8 | **239.5** / 252.2 | 213.7 / 241.2 | 247.2 / 260.0 | 389.9† |
| `tok/step` | 3.63-4.00 | 3.43-3.78 | 3.65-3.93 | — |
| meanTTFT C1 | 186-194 ms | 175-187 ms | 179-183 ms | — |

† the published C8 row is from a different box. All three checkpoints land within
±10% of each other and of the published DFlash2 single profile: the two uncensored
fine-tunes decode at the base model's speed.

### Batch profile (`KV=fp8`, `MAX_SEQS=64`, `MAX_LEN=150000`, int8 activations, no speculation)

| row | orca | swift | autoround-base | published (batch) |
|---|---|---|---|---|
| 64conc 128in/512out | 858.2 | 787.0 | 781.1 | ~1,035 |
| 64conc 256in/256out | 608.1 | 571.7 | 568.9 | — |
| cohort C1 / C2 / C4 / C8 (T=default) | 40.9 / 70.3 / 143.0 / 272.3 | 40.8 / 72.7 / 143.0 / 273.0 | 40.8 / 73.2 / 141.7 / 269.1 | 45.5 / 86.3 / 168.3 / 324.9 |
| prefill 1024 (c1/c4/c16) | 1498 / 1574 / 1637 | 1503 / 1582 / 1633 | 1489 / 1570 / 1629 | — |
| prefill 4096 (c1/c4/c16) | 1541 / 1576 / 1566 | 1547 / 1578 / 1569 | 1536 / 1575 / 1564 | — |
| prefill 16384 (c1/c4/c8) | 1429 / 1438 / 1430 | 1427 / 1439 / 1430 | 1425 / 1434 / 1428 | — |
| prefill 65536 (c1/c2) | 1051 / 1052 | 1050 / 1052 | 1051 / 1051 | — |
| prefill 102400 (c1) | 884 | 883 | 883 | — |
| 1x100k/256 | TTFT 111.8 s, TPOT 30.0 ms | TTFT 111.8 s, TPOT 30.1 ms | TTFT 112.0 s, TPOT 30.1 ms | — |
| 4x60k/1024 conc4 | 14.21 tok/s, medITL 36.2 ms, 288 s | 14.15 tok/s, medITL 32.2 ms, 289 s | 14.15 tok/s, medITL 32.3 ms, 289 s | — |

Batch reads ~17% below the published 64-concurrency row and ~16% below the
published cohorts, consistently across all three checkpoints — the box's pinned
`GPU_UTIL=0.93` (published rows used the batch profile default 0.972, i.e. a
larger KV pool) plus the 250 W cap, not a checkpoint difference: every batch row
agrees across the three checkpoints to within 1-3%.

`4x60k/1024 conc4` is KV-bound, not compute-bound: 4 x 61k = 244k resident tokens
against a ~130k-token fp8 pool, so vLLM preempts and recomputes (289 s for 4096
output tokens). Identical on all three, so it compares cleanly.

### Quality gate (`bench/quality_battery.py`, single config)

| checkpoint | PPL en / da / code / **all** (32,679 tok) | GSM8K (n=200, greedy) |
|---|---|---|
| orca | 10.7994 / 10.9052 / 3.2686 / **8.2168** | 0.945 (387 tok) |
| swift | 10.8014 / 10.9196 / 3.3368 / **8.2614** | **0.980** (356 tok) |
| autoround-base | 10.6784 / 10.8476 / 3.3017 / **8.1858** | 0.950 (377 tok) |
| published (base model) | — / — / — / 8.09 (33k tok) | 96.5% |

All three pass; the two uncensored checkpoints are within 1% of the base model's
perplexity, and swift is the strongest on GSM8K.

mhenrichsen:
@sudoeste is right, and it is checkable without downloading the 19 GB — the safetensors headers say it directly. Reading just the header of each shard over an HTTP range request:

```
model_extra_tensors.safetensors    27 keys, all mtp.*   mtp.fc.weight_packed  I32 (5120, 1280)
model-00004-of-00004.safetensors  267 keys, 34 mtp.*    mtp.fc.weight_packed  I32 (5120, 2560)
```

`5120 x 1280` is the int4 packing of a 10240-wide input (`10240/32*4`), `5120 x 2560` is the int8 one (`10240/32*8`). Both are physically present, under the same key. And `model.safetensors.index.json` routes only three keys to the extras file —

```
mtp.draft_lm_head.weight_packed / .weight_scale / .weight_shape
```

— so every other `mtp.*` key the index names still points at the int8 copies in shard 4, while the int4 recalibration sits in the extras file unreferenced. vLLM loads every key in a file it opens, so it sees both widths for `mtp.fc.weight_packed` and dies on a shape assert; that is the exact stale-tensor collision [#139](https://github.com/syv-ai/HyperQwen/pull/139) documents as a build note ("strip the superseded int8 MTP tensors from any shard you keep").

Fix on the HF side is one of: strip the 24 superseded `mtp.*` tensors (the ones duplicated in extras) from `model-00004-of-00004.safetensors` and repoint the index at the extras copies, or drop the int4 MTP from extras and keep the int8 layout consistently. Either way the index and the shard contents have to agree.

Holding the docs entry until then — an entry in `docs/third-party-checkpoints.md` says "this serves with no preparation", and right now it does not. Ping here when the repo is fixed and I will verify the headers again and merge; the write-up itself is good, and the honest 31.2%-vs-claimed-58.3% reasoning-token number is exactly the kind of thing worth having in there.

Separately, the two throughput rows (`CTX=fast`-equivalent 148 tok/s, `CTX=long`-equivalent 96) — what produced those? Every other number in that file is `bench/run_benchmarks.sh`, and 148 at C1 is well above what this stack does on a 3090, so it reads as a different measurement. If it was your own client, say so in the entry the way the leminkozey one does.

ultimaterex:
@mhenrichsen Fixed: https://huggingface.co/ultimaterex/Swift-Qwen3.8-27B-Uncensored-W4A16-AutoRound/commit/29c95a9c5a0b36246aae41b45c2e6732d9db48e0

Went with your first option. Stripped the 27 superseded int8 `mtp.*` tensors from shard 4 (your 24, plus the 3 `mtp.draft_lm_head.*` keys shard 4 also still held copies of) and pointed all 27 index entries at `model_extra_tensors.safetensors`. Shard 4 goes from 3.25 to 2.61 GB. The 240 remaining tensors are byte-identical, and shards 1-3 and the extras file are untouched.

Re-checked the headers on the live repo: 2,022 keys, none in more than one file, index matches contents, `mtp.fc.weight_packed` exists once at (5120, 1280). Booted vLLM 0.28.0 with `SPEC=mtp`: the previous upload crashes at load with the shape mismatch you found; the fixed one boots and serves.

Also ran `bench/quality_battery.py` on CTX=huge (SPEC=dflash2, so unaffected by gotcha 46) to make sure the tensor fix didn't regress anything: PPL en 11.22 / da 11.49 / code 3.50, GSM8K 200q 95.5% acc. In line with the other checkpoints in `bench/quality-data/`.

Pushed the updated entry to the PR branch.

TyroneNel:
Verified the fix on the live files without downloading the 19 GB — safetensors headers plus the index over HTTP ranges, then this repo's own gate against a local copy.

**The collision is gone.** 2022 index entries / 2022 keys across the five safetensors files, 0 missing, 0 duplicated, 0 orphans; `mtp.fc.weight_packed` now resolves only to `model_extra_tensors.safetensors`; shard 4 is down to 240 keys and its only remaining `mtp.*` are the 7 BF16 norms (`mtp.norm`, `pre_fc_norm_embedding`, `pre_fc_norm_hidden`, `layers.0.{input,post_attention}_layernorm`, `q_norm`, `k_norm`) — no packed `mtp.*` left in it.

**The repo's pre-serve gate passes.** `bash verify.sh --no-server MODEL=…` → model checks 7/7 PASS, including the cross-shard duplicate check (`every index tensor lives in exactly its mapped shard`) that exists in `main` precisely for this failure mode. Since the entry's claim is "serves without any preparation", a line telling readers to run that command is the cheapest way to make it checkable — the leminkozey entry has one.

**One factual correction before merge.** The entry says the int4-GPTQ requant covers `lm_head`, `embed_tokens` and the MTP module. `embed_tokens` is int8 in the artifact: `model.language_model.embed_tokens.weight_packed` is `I32 [248320, 1280]` with a **BF16** scale (`group_2`, `num_bits: 8`) — int4 would be 640 columns. That matches the pipeline too: `prepare/quant_embed.py` hardcodes `BITS = 8`, `quant_heads_stream.py` has `HEAD_BITS = 8` (a constant, not a flag — only `--mtp-bits` is exposed), and `drafter/gptq_lm_head.py` rewrites `lm_head` only. Suggested wording: "int4-GPTQ `lm_head` and MTP module, `embed_tokens` kept int8".

**One provenance point.** The entry calls the draft vocabulary "a 40,960-token draft vocabulary build". `mtp_draft_vocab_ids.pt` is set-identical to this repo's shipped `prepare/draft_vocab_ids.json` (40,960 ids, compared as sets) — it is the base list, not one counted over this model's outputs. Worth stating either way, because the own-output rebuild is the stronger result for this finetune: our `TyroneNel/Swift-Qwen3.8-27B-Uncensored-W4A16` card measures 98.90% held-out coverage vs 98.08% for the base list, and #139's build ships 25,879 rows at 99.8% with the acceptance gain. "ships the repo's base list" reads better than "build", and costs nothing.

Minor: `bench/quality-data/` is gitignored (0 tracked files), so that citation isn't resolvable from the repo for a reader — `docs/benchmarks.md` (GSM8K 96.5%) and `docs/long-context.md` (95.0–97.0%) are.

FWIW the 31.2%-vs-58.3% reasoning-token number reads well next to the adapter's headline claim — keeping the shortfall visible with the per-task caveat is the right call.

mhenrichsen:
Merging this.

**The artifact is fixed and two people checked it independently.** @TyroneNel walked the live files over HTTP ranges — 2,022 index entries against 2,022 keys, nothing duplicated or orphaned, `mtp.fc.weight_packed` resolving only to the extras file, shard 4 down to 240 keys with no packed `mtp.*` left — and ran this repo's own gate, `verify.sh --no-server`, to 7/7 including the cross-shard duplicate check that exists for exactly this failure. I had checked the broken state the same way; the fix holds from both directions, and you booted `SPEC=mtp` on it.

**I pushed one commit to your branch** with the two corrections from the review, both checkable without the 19 GB:

- `embed_tokens` is **int8**, not int4. The pipeline says so — `prepare/quant_embed.py` has `BITS = 8` and `prepare/quant_heads_stream.py` has `HEAD_BITS = 8` (a constant; only `--mtp-bits` is exposed) — and so does your artifact: `config.json` `group_2` is `num_bits: 8` and the pack is `I32 [248320, 1280]`, where int4 over a 5120-wide hidden would be 640 columns. `lm_head` and the MTP module are genuinely int4. The entry now reads "int4-GPTQ `lm_head` and the MTP module, `embed_tokens` kept int8", with the reason inline.
- The draft vocabulary is **this repo's shipped 40,960-id list**, not a build over this checkpoint's own outputs. Worth stating because the own-output rebuild is the stronger move for a finetune and someone reading "a 40,960-token draft vocabulary build" would think it had already been done — #139 measured Swift emitting ~25.9k distinct tokens against base Qwen's ~54k, so a rebuilt head is 25,879 rows at 99.8% coverage and recovers the acceptance deficit. That is the obvious next improvement to this checkpoint, and now the entry points at it.

Also swapped the `bench/quality-data/` citation for `docs/benchmarks.md` and `docs/long-context.md` (the former is gitignored, so a reader cannot resolve it), and added the `verify.sh --no-server` line — since the entry's claim is that it serves with no preparation, the cheapest way to make that checkable is to tell people the command.

Nothing else needed from you. Dropping the unverifiable 148/96 tok/s figures rather than defending them was the right call, and the 31.2%-against-58.3% reasoning-token result with the per-task caveat is more useful than the adapter's headline number.

PR #157, opened 2026-09-19
bench: don't abort the run when spec-decode metrics are absent (batch mode)

TyroneNel:
Two defects in one line of `spec()`, both hit while benchmarking three checkpoints on an RTX 3090.

**1. Batch mode cannot complete.** Batch runs with `speculative_config=None`, so vLLM registers no `vllm:spec_decode_num_{drafts,accepted_tokens}_total` series. `spec()`'s `grep` then exits 1 and, under `set -e` + `pipefail`, the assignment `S0=$(spec)` aborts the whole run at the first cohort - silently: no `FAIL`/`INVALID` line, no cohort log, exit status 1. `bash bench/run_benchmarks.sh batch` therefore never reaches the prefill/long matrices its own header documents (`batch --prefill`, `batch --long`). Fix: `return 0`, so absence is an empty sample - which `tokstep()` already renders as `-`.

**2. `tok/step` was reading the label set.** The series are labelled, e.g. `vllm:spec_decode_num_drafts_total{engine="0",model_name="qwen3.8-27b"} 123.0`, so field 2 is `{engine="0",...}` and only the last field is the sample. `$NF` instead of `$2`. (Happy to drop this hunk if you would rather keep the PR to the abort fix.)

Repro (batch profile, any checkpoint):
- before: two `ROW 64conc` lines, then exit 1, no `cohort_Tdefault_c1.log`
- after: cohorts run with `tok/step=-`, then the prefill matrix and `RESULT PASS`

Evidence: engine log during the run showed `Running: 1 reqs, Waiting: 14 reqs` at 17% KV for the single-profile attempt (the matrices need the batch profile's 64 seats and `--max-num-batched-tokens 2048`); the aborted batch run's directory contained only `warmup.log`, `batch_128_512.log`, `batch_256_256.log`, `manifest.json`, `prompts_real.jsonl`.

Single mode is unaffected either way: the counters exist there, so every printed value is unchanged.

Note for route A (`docker compose exec ... bash bench/run_benchmarks.sh batch`): the harness runs from the image's copy of `bench/`, so this only takes effect there after an image rebuild (or by bind-mounting `bench/` over `/app/bench`). The single-mode route used by the README field-test section is unaffected.


mhenrichsen:
Thanks for the patch, but I can't reproduce the failure it describes, and I think the PR is written against a different harness than the one in this repo.

Checked at `feaffb6` (`bench/run_benchmarks.sh`):

- **No `set -e`, no `pipefail`.** Line 14 is `set -u` and that is the only `set` in the file (`grep -n 'set -' bench/run_benchmarks.sh` → line 14 `set -u`, line 56 `set -- $W`). A failing `$(spec)` therefore assigns the empty string and the script continues: `bash -c 'set -u; spec() { echo x | grep -E "^nope" | awk "{print \$NF}"; }; S0=$(spec); echo "survived [$S0]"'` prints `survived []`. `tokstep()` then renders `-`, which is the behaviour your patch is aiming for — it is already the behaviour.
- **The artefacts you list don't exist here.** `grep -rn "manifest.json\|RESULT PASS\|warmup.log"` over the repo returns nothing: this harness writes no `manifest.json`, no `warmup.log`, and prints no `RESULT PASS` / `FAIL` / `INVALID` lines. The warmup on line 54 goes to `/dev/null`, and `prompts_real.jsonl` lives in `bench/`, not in `$OUT`.
- **`$2` is already the sample.** Prometheus emits `vllm:spec_decode_num_drafts_total{engine="0",model_name="qwen3.8-27b"} 123.0` with no space before the label set, so `$2` and `$NF` are the same field. `$NF` is the more defensive choice and `bench/real_rep.sh:13` already uses it — but the stated reason (field 2 being `{engine="0",...}`) doesn't hold, and if it did hold, single mode would have been printing garbage `tok/step` all along rather than the 3.2-3.4 everyone reports.

So: what would actually help is the real failure. Can you post the exact invocation and the last ~20 lines of output from the aborted batch run, plus `git rev-parse HEAD` in the tree you ran it from? If you invoked it as `bash -e bench/run_benchmarks.sh batch`, or from a wrapper that sets `-e -o pipefail`, then the abort is real under that invocation and the `return 0` is worth having — I'd just want the commit message to say that rather than describe a `set -e` this file doesn't have.

The one thing `real_rep.sh` does that both `run_benchmarks.sh` and `prefill_ab.sh` skip is `grep -v created`. If your prometheus client emits `..._total_created` alongside the counters, the unanchored `^vllm:spec_decode_num_(drafts|accepted_tokens)_total` regex picks up four samples instead of two and `tokstep()` reads the wrong pair — that *is* a way to get nonsense out of `spec()`, and it would be a real fix in both files. Worth checking `curl -s localhost:18020/metrics | grep spec_decode_num_drafts` on your server.

TyroneNel:
You are right on all three points, and I owe you a clear retraction rather than a defence.

**The abort is mine, not this harness's.** I ran the batch profile through a local wrapper of my own that sets `set -euo pipefail`, and `bench/run_benchmarks.sh` on `main` has only `set -u` on line 14. Under the invocation this repo documents, a failing `$(spec)` assigns the empty string, the run continues, and `tokstep()` renders `-` — exactly as you showed. The artefacts I listed (`manifest.json`, `warmup.log`, `RESULT PASS`) are from that wrapper too, which is why `grep -rn` finds none of them here. The PR describes a harness that is not the one in this repo, and the commit message asserts a `set -e` this file does not have.

**`$2` is already the sample.** Prometheus emits no space before the label set, so `$2` and `$NF` are the same field, and my stated reason for the change was simply wrong.

**And the `_created` hypothesis does not fire either** — worth writing down so you do not spend time on it. vLLM registers the counter as `vllm:spec_decode_num_drafts` with no `_total` suffix (`v1/spec_decode/metrics.py`, `counter_specs`). `prometheus_client` builds sample names as `name + suffix` (`metrics.py:92`), so the exposition carries:

```
vllm:spec_decode_num_drafts_total{...} 123.0
vllm:spec_decode_num_drafts_created{...} 1.7e+09
```

The scrapers' regex is anchored as `^vllm:spec_decode_num_(drafts|accepted_tokens)_total`, and `..._created` cannot match it. `_total_created` would only appear if the metric were *declared* with `_total` in the name (`_build_full_name` strips it for counters, `metrics.py:34`) — vLLM does not. So `grep -v created` in `bench/real_rep.sh` is inert today, and adding it to `run_benchmarks.sh` and `prefill_ab.sh` would be a no-op. I had it queued as a follow-up PR and dropped it rather than send you a change that cannot do anything.

Closing this. Thank you for checking it against the actual file at `feaffb6` instead of taking the description at face value — the three counter-measurements are the reason I went and looked at my own wrapper.

PR #158, opened 2026-09-19
verify: reject asymmetric pack-quantized groups in the model checks

TyroneNel:
## What

`verify.sh`'s model checks now FAIL if any `pack-quantized` group in `config.json`'s `quantization_config.config_groups` declares asymmetric weights (`symmetric` is not `true`). Symmetric groups produce one PASS line. The FAIL names the offending group, its targets, and the fix (`prepare/quant_heads_stream.py`).

## Why

The packed tensors `prepare/` writes are symmetric (no zero point), but a foreign export can declare asymmetric groups over the same pack-quantized layout. vLLM then looks for `weight_zero_point` tensors that were never written and dies far from the cause. This is the consumer-side twin of the asymmetric-AWQ-body case that `prepare/quant_heads_stream.py` already normalizes away on the producer side (it forces `symmetric=true / zp_dtype=null` on every group it writes, lines 246-251), and it is exactly the failure field-reported in [#139 (comment)](https://github.com/syv-ai/HyperQwen/pull/139#issuecomment-5735898343):

> Tried your model but can't use dstyle = int8 — Oh, I see it's asymmetric :/

`verify.sh --no-server` is the documented pre-serve gate for third-party checkpoints (`docs/third-party-checkpoints.md`), so this turns a cryptic boot-time abort into a named, actionable FAIL before a serve attempt is ever made.

## Verification

- Real checkpoints (both fully symmetric): orca `Qwen3.8-27B-Uncensored-W4A16` and Swift `Swift-Qwen3.8-27B-Uncensored-W4A16` → `PASS all 4 quant groups symmetric`, all other model checks unchanged.
- Negative control: injected `symmetric: false` + `zp_dtype: torch.int8` into a copy of the Swift config → `FAIL quant group group_0 (['Linear']) declares asymmetric weights (zero-point) ...` with a nonzero exit.
- End-to-end in the container: image rebuilt from this branch, `verify.sh --install` inside the container → `verify: OK (0 failures)`; the model-check block run inside the container against the mounted models passes/fails as above.
- `bash -n verify.sh` clean; all patch layers rebuilt from the same `patches/series` (44 patches PASS at fuzz 0).

Single file, +13 lines, no behavior change for any checkpoint the repo already serves.


mhenrichsen:
The idea is right — a checkpoint that declares zero points over tensors written without them should be caught before a serve attempt — but as written this check fails the repo's own documented worked example.

**The false positive.** `prepare/quant_heads_stream.py` normalizes `symmetric`/`zp_dtype` only on the groups it *writes* (`group()`, lines 219-230, used for `group_1`/`group_2`/`group_3` — lm_head, embed_tokens, mtp). It deliberately leaves `group_0` alone, and says why at lines 224-225: *"tensors written here are symmetric with no zero point, regardless of what the body group uses (AWQ bodies are asymmetric)"*. The body keeps its zero points because the body tensors actually have them.

The AWQ worked example this repo ships — `philbert440/Qwen3.8-27B-Uncensored-Aggressive-W4A16-AWQ`, the default in `prepare/fetch_thirdparty.py` and an entry in `docs/third-party-checkpoints.md` — has exactly that shape. Straight from its published config:

```
philbert440/...-AWQ            group_0  format=pack-quantized  symmetric=False  bits=4  targets=['Linear']
dbirks/Qwen3.8-27B-W4A16-AutoRound  group_0  format=pack-quantized  symmetric=True   bits=4  targets=['Linear']
```

After `prepare/quant_heads_stream.py` that first config still carries `group_0` asymmetric, plus three symmetric head groups. Your check would print

```
FAIL quant group group_0 (['Linear']) declares asymmetric weights (zero-point): ... Requantize with prepare/quant_heads_stream.py
```

on a checkpoint that serves — and the remedy it names is the script that produced that very config. Your test matrix (orca, Swift) is symmetric end to end, so it never touched this path.

**What I'd merge.** Scope the check to the groups `prepare/` writes, i.e. the ones whose `targets` are the head patterns (`re:.*lm_head$`, `re:.*embed_tokens$`, `re:^mtp\..*` / `re:^mtp\.layers\..*`) rather than every `pack-quantized` group. Those are the tensors we know were written without zero points, and a foreign export declaring zero points over them is the failure you're after. A body group with `symmetric=false` is legitimate here and should stay silent — or at most `warn`.

One smaller thing while you're in there: `g.get("format") == "pack-quantized"` works on both configs I checked (the per-group `format` is populated), but compressed-tensors also allows a group to leave `format` null and inherit `quantization_config["format"]`. Falling back to the top-level value costs one line and removes the chance of the check going vacuously green.

Push a follow-up commit and I will take it — the verify.sh plumbing and the message wording are both fine, it's only the selector that is too wide.

TyroneNel:
Scoped exactly as you asked — pushed as `2d8fd6e` (on top of `6a5fbf0`).

- The selector is now the head patterns intersected with each group's `targets` (`re:.*lm_head$`, `re:.*embed_tokens$`, `re:^mtp\..*`), so a body group with `symmetric=false` stays silent. Your philbert440 case now PASSes; before this commit it printed `FAIL quant group group_0 (['Linear']) ... Requantize with prepare/quant_heads_stream.py`.
- `format` falls back to `quantization_config["format"]` when the group leaves it null. I reproduced your vacuous-green case: a head group with `format: null` and zero points printed `PASS all 4 quant groups symmetric` before, and FAILs now.
- Both columns measured (`verify.sh` model section), four-case synthetic matrix plus the real dirs:

| case | before `6a5fbf0` | after `2d8fd6e` |
|---|---|---|
| four symmetric groups (orca g128, 67 shards; Swift g128, 67 shards; `ultimaterex/Swift-...-AutoRound`, 5 shards) | PASS 7/7 | PASS 7/7 |
| body group asymmetric (`group_0` `symmetric=false` + `zp_dtype`, philbert440 shape) | **FAIL** | PASS |
| head groups asymmetric (`group_1`/`group_2`/`group_3` with `zp_dtype`) | FAIL | FAIL, one line per head group |
| head group asymmetric with group `format: null` (inherits the top-level format) | **PASS (vacuously green)** | FAIL |

`bash -n` clean, and the branch merges clean against current `main`. The matrix is four header-only dirs — no weights, under a second per case — so if you would rather have the selector pinned by a test than by this comment, I can contribute it as `bench/test_verify_zp_selector.py`.

mhenrichsen:
Scoped the way I asked, and I verified it on real configs rather than on the reasoning — including two checkpoints that merged today, so this check will meet them immediately.

Ran your block against five community exports plus the shipped one, and against the pre-narrowing version for contrast:

| checkpoint | group_0 `symmetric` | head groups | narrowed | pre-narrowing |
|---|---|---|---|---|
| dbirks/…-W4A16-AutoRound (reference) | true | — | PASS | pass |
| philbert440/…-Aggressive-W4A16-AWQ | **false** | — | **PASS** | FAIL |
| liamwh/Swift-…-syv-fast (#139) | **false** | all true | **PASS** | FAIL |
| ultimaterex/Swift-…-Uncensored (#144) | true | all true | PASS | pass |
| TyroneNel/…-Uncensored-W4A16-g128 | true | all true | PASS | pass |

The liamwh row is the one that settles it: an AWQ body with `symmetric=false` *and* a full set of rebuilt head groups, i.e. exactly the shape the old selector could not tell apart from a broken export. It merged an hour ago, so the narrowing is not hypothetical — without it, `verify.sh` would fail on a checkpoint this repo documents as serving with no preparation.

And the check still bites when it should: I flipped `group_1` (`re:.*lm_head$`) to `symmetric: false` on that same config and it fails with your message.

**One edge to fix before I merge, found the same way.** `g["weights"].get("symmetric") is not True` fails a head group that *omits* the key, and omitting it means symmetric: `compressed_tensors.quantization.quant_args.QuantizationArgs` declares `symmetric: bool = True` (quant_args.py:190). Our own `prepare/quant_heads_stream.py` always writes the key explicitly, so this can only bite a foreign export — which is precisely the population this check exists for. Synthetic proof, same liamwh config with the key deleted from `group_2`:

```
  FAIL  head group group_2 (['re:.*embed_tokens$']) declares asymmetric weights (zero-point)
```

`g["weights"].get("symmetric", True) is not True` closes it. Same for the `zp_dtype` side if you ever extend this: absent is not asymmetric.

The `format` fallback to `quantization_config["format"]` is right and worth keeping — three of the five configs above set it per group *and* at the top level, and a group that leaves it null is legal.

Push that one-word change and I will merge.

TyroneNel:
`g["weights"].get("symmetric", True)` pushed. Thank you for finding that with a real config rather than by reading the diff — an omitted key is exactly the shape a foreign export arrives in, and I had written the check so that the one population it exists for would trip it.

Exercised by running the block out of `verify.sh` itself against synthetic configs:

| case | before | after |
|---|---|---|
| head group omits `symmetric` (your liamwh config, key deleted from `group_2`) | **FAIL** | **PASS** |
| head group `symmetric: false` | FAIL | FAIL (unchanged) |
| AWQ body asymmetric, heads symmetric (philbert440) | PASS | PASS (unchanged) |

Noted on `zp_dtype` too: absent is not asymmetric, so if that side is ever added it gets the same defaulting. The `format` fallback is untouched.

Branch is `MERGEABLE / CLEAN` against current `main`.

mhenrichsen:
Merged. Re-ran the block exactly as it ships in `verify.sh` (extracted from the heredoc, not retyped) against the same seven configs:

```
dbirks  AutoRound (reference)            PASS
liamwh  Swift syv-fast (AWQ body)        PASS
philbert440 AWQ (asymmetric body)        PASS
TyroneNel Uncensored g128                PASS
ultimaterex Swift+Uncensored             PASS
synthetic: lm_head group symmetric=false FAIL   <- the case it exists for
synthetic: embed group omits symmetric   PASS   <- was FAIL before this push
```

The `QuantizationArgs` citation in the comment is the right thing to have put there — the next person tempted to tighten `.get("symmetric")` back to strict now has the reason in front of them. Thanks for turning this round the same evening.

